PR #6, opened 2016-12-09
Removing the QUOTED_IDENTIFIERS_IGNORE_CASE parameter in the default …

etduwx:
…connection settings.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -84,7 +84,14 @@ private[snowflake] class JDBCWrapper {
         val isSigned = rsmd.isSigned(i + 1)
         val nullable = rsmd.isNullable(i + 1) != ResultSetMetaData.columnNoNulls
         val columnType = getCatalystType(dataType, fieldSize, fieldScale, isSigned)
-        fields(i) = StructField(columnName, columnType, nullable)
+        fields(i) =
+          StructField(
+            if(columnName == columnName.toUpperCase) // Add quotes around column names if not all uppercase
+              columnName
+            else
+              s""""$columnName"""",
+            columnType,
+            nullable)
         i = i + 1
       }
       new StructType(fields)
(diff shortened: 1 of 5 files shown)


etduwx:
@MarcinZukowski  do you know why Travis is failing with the auth issue?  All of those tests pass when I run them locally.

etduwx:
@MarcinZukowski 

Nevermind, weird. I manually restarted the build and now it passed. A transient failure, I guess. :-)

smtakeda:
Hi. I think Travis CI changed the IP ranges. We restrict access to the test account by IP. We'll reconfigure it.

-- Shige Takeda

etduwx:
@smtakeda Thanks.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -84,7 +84,14 @@ private[snowflake] class JDBCWrapper {
         val isSigned = rsmd.isSigned(i + 1)
         val nullable = rsmd.isNullable(i + 1) != ResultSetMetaData.columnNoNulls
         val columnType = getCatalystType(dataType, fieldSize, fieldScale, isSigned)
-        fields(i) = StructField(columnName, columnType, nullable)
+        fields(i) =
+          StructField(
+            if(columnName == columnName.toUpperCase) // Add quotes around column names if not all uppercase
```
MarcinZukowski: Don't we need something for names with special characters (space, quote, doublequote etc) ?

We should test columns like that.

etduwx: Good point. I now pattern match for all identifiers (after upcasting) allowed by Snowflake without quotes (as defined here: https://docs.snowflake.net/manuals/sql-reference/identifiers.html), and then add quotes to names that don't match during DataFrame column resolution. The user should now have to manually quote (in Spark) all identifiers where they would normally have to when directly using SQL against Snowflake.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -206,4 +208,8 @@ private[snowflake] case class SnowflakeRelation(
     val fieldMap = Map(schema.fields.map(x => x.name -> x): _*)
     new StructType(columns.map(name => fieldMap(name)))
   }
+
+  private def isQuoted(name: String): Boolean = {
+    name.startsWith("\"") && name.endsWith("\"")
```
etduwx: This should never happen since we always insert quotes around special names for DataFrame columns. If the user makes a mistake, then Spark would error-out during name resolution before the code reaches this point.

PR #10, opened 2017-04-22
Fixed issue with objects as value in java.util.properties

shashankgowdal:
java.util.Properties().getProperty returns null when the value is an object other than String. This is resulting in a runtime exception when SparkSnowflake is used in an application where other JDBC drivers co-exist.
We are facing the issue with Postgres driver which in the connect method tries to read the properties using getProperty method.
https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/Driver.java#L221


etduwx:
Thanks @shashankgowdal .

Issue #12, opened 2017-05-12
table creation fails if column name is a keyword


jhuangtw-dev:
calling `dataframe.write()` to a table that does not exist yet results in:
```
CREATE TABLE IF NOT EXISTS test_schema.test_table (id INTEGER NOT NULL, created_at TIMESTAMP, updated_at TIMESTAMP, name STRING NOT NULL, order INTEGER)
```
which fails with: 
```
SQL compilation error: syntax error line 1 at position 138 unexpected ''order''.
```

whereas:
```
CREATE TABLE IF NOT EXISTS test_schema.test_table (id INTEGER NOT NULL, created_at TIMESTAMP, updated_at TIMESTAMP, name STRING NOT NULL, "order" INTEGER)
```
succeeds.

would it make sense to always double-quote the column names during that SQL statement construction?

etduwx:
Thanks, @jhuangtw-dev . Will accept a PR on this; otherwise, I will fix it at my next opportunity.

mckim0928:
Hi @etduwx, I was hopping to take a stab at this PR, just started looking inside the codebase, and was wondering if you could point me in the right direction to get the fix in.

mckim0928:
@etduwx, created a PR with my fix, but having a little issue with setting up my travis credential. Would really appreciate your guidance!

etduwx:
Thanks @mckim0928 ! I've merged your PR and it will be included in our next release, which I am hoping will be very soon. @jhuangtw-dev , when the new version comes out, please reopen an issue if it is still not resolved.

PR #13, opened 2017-06-08
Double quoting columns create sql

mckim0928:
PR regarding issue #12 

calling dataframe.write() to a table that does not exist yet results in:

```
CREATE TABLE IF NOT EXISTS test_schema.test_table (id INTEGER NOT NULL, created_at TIMESTAMP, updated_at TIMESTAMP, name STRING NOT NULL, order INTEGER)
```
which fails with:

```
SQL compilation error: syntax error line 1 at position 138 unexpected ''order''.
```
whereas:

```
CREATE TABLE IF NOT EXISTS test_schema.test_table (id INTEGER NOT NULL, created_at TIMESTAMP, updated_at TIMESTAMP, name STRING NOT NULL, "order" INTEGER)
```
succeeds.

In the JDBCWrapper class, modified the schemaString method so that column names are always double-quoted in the createTableSql method in the SnowflakeWriter class. 

Added in a test case inside SnowflakeSourceSuite, which might be redundant with a modified, pre-existing test case.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -182,7 +182,7 @@ private[snowflake] class JDBCWrapper {
     schema.fields.foreach { field =>
       {
         val name = field.name
-       
+        val formattedName = if (isQuoted(name)) name else quotedName(name)
         val typ: String = field.dataType match {
           case IntegerType => "INTEGER"
           case LongType => "INTEGER"
@@ -205,7 +205,7 @@ private[snowflake] class JDBCWrapper {
             throw new IllegalArgumentException(s"Don't know how to save $field of type ${field.name} to Snowflake")
         }
         val nullable = if (field.nullable) "" else "NOT NULL"
-        sb.append(s""", ${if (isQuoted(name)) name else quotedName(name)} $typ $nullable""".trim)
+        sb.append(s""", ${formattedName.replace("\"", "\\\"")} $typ $nullable""".trim)
       }
     }
     if (sb.length < 2) "" else sb.substring(2)
@@ -215,10 +215,12 @@ private[snowflake] class JDBCWrapper {
     name.startsWith("\"") && name.endsWith("\"") 
   }
   private def quotedName(name: String): String = {
-      if (name.matches("[_A-Z]([_0-9A-Z])*"))
+    // Name legality check going from spark => SF. 
+    // If the input identifier is legal, uppercase before wrapping it with double quotes.
+      if (name.matches("[_a-zA-Z]([_0-9a-zA-Z])*"))
         "\"" + name.toUpperCase + "\""
       else
-        name
+        "\"" + name + "\""
   }
 
   /**
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -182,7 +182,7 @@ private[snowflake] class JDBCWrapper {
     schema.fields.foreach { field =>
       {
         val name = field.name
-        val quotedName = if (isQuoted(name)) name else "\"" + name + "\""
+       
         val typ: String = field.dataType match {
           case IntegerType => "INTEGER"
           case LongType => "INTEGER"
@@ -205,7 +205,7 @@ private[snowflake] class JDBCWrapper {
             throw new IllegalArgumentException(s"Don't know how to save $field of type ${field.name} to Snowflake")
         }
         val nullable = if (field.nullable) "" else "NOT NULL"
-        sb.append(s""", $quotedName $typ $nullable""".trim)
+        sb.append(s""", ${if (isQuoted(name)) name else quotedName(name)} $typ $nullable""".trim)
       }
     }
     if (sb.length < 2) "" else sb.substring(2)
@@ -213,6 +213,12 @@ private[snowflake] class JDBCWrapper {
 
   private def isQuoted(name: String): Boolean = {
     name.startsWith("\"") && name.endsWith("\"") 
+  }
+  private def quotedName(name: String): String = {
+      if (name.matches("[_A-Z]([_0-9A-Z])*"))
+        "\"" + name.toUpperCase + "\""
+      else
+        name
   }
 
   /**
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -182,6 +182,7 @@ private[snowflake] class JDBCWrapper {
     schema.fields.foreach { field =>
       {
         val name = field.name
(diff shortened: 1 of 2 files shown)


etduwx:
@mckim0928 Thanks for the PR! Don't worry about the Travis credentials; PRs will naturally fail the integration tests due to different accounts. However, if you have access to an SF account you can run the unit tests manually following the .conf file example in the repo. :)

Just one comment: In Snowflake, unquoted identifiers automatically equate to the all-caps version of the string, e.g., 

`a` is the same as `"A"` but different from `"a"`. We use a regex pattern also in `SnowflakeJDBCWrapper.scala:89` to test for this for the SF-> Spark case. 

For this, maybe we can check for quotes, and if it needs quoting, we can do the same regex match check -- if it matches, we promote to all caps before wrapping it in quotes, otherwise we leave the original string as-is.

mckim0928:
@etduwx  Thanks so much for the feedback! That totally makes sense. I will soon push those suggested modifications. Current unit tests are passing (only ran with sbt though) except for the 4 cases that require input paths, but I will try running manual tests with Handshake's SF account credentials if i can! Thank you again for taking the time to review and comment :)

mckim0928:
@etduwx Just added the suggested logic, but wasn't sure how exactly to test this...also excluded the `${name.replace("\"", "\\\"")}` for now (see above), but happy to modify that / add in more test cases where you think appropriate!

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -214,6 +214,12 @@ private[snowflake] class JDBCWrapper {
   private def isQuoted(name: String): Boolean = {
     name.startsWith("\"") && name.endsWith("\"") 
   }
+  private def quotedName(name: String): String = {
+      if (name.matches("[_A-Z]([_0-9A-Z])*"))
+        "\"" + name.toUpperCase + "\""
+      else
+        name
```
etduwx: I think we still want to add quotes here...we always quote identifiers now, right? The `if` only checks whether we promote to upper-case.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -214,6 +214,12 @@ private[snowflake] class JDBCWrapper {
   private def isQuoted(name: String): Boolean = {
     name.startsWith("\"") && name.endsWith("\"") 
   }
+  private def quotedName(name: String): String = {
+      if (name.matches("[_A-Z]([_0-9A-Z])*"))
+        "\"" + name.toUpperCase + "\""
```
etduwx: Woops, I think I messed up the logic here. We shouldn't use the same regex because this is going the other direction (for the Snowflake-> Spark case, we perform the check on names for which the upper-case promotion has already been performed, but here we start from any user-given column names). 

When starting from Spark we instead need to check every unquoted identifier (here in `quotedNames`) to see if it contains any characters that wouldn't otherwise be legal without being quoted. If so, then we simply add the quotes. If it's all legal, then we upper-case and quote. So we need another regex check for name legality. I think this basically amounts to changing the pattern to also include lowercase letters.

See here:
https://docs.snowflake.net/manuals/sql-reference/identifiers.html

> Identifiers that are not enclosed by double quotes are case-insensitive and must start with a letter or the underscore (_) character. In addition, they can only contain a combination of alphanumeric characters and the underscore character.

I hope this makes sense.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -204,12 +205,22 @@ private[snowflake] class JDBCWrapper {
             throw new IllegalArgumentException(s"Don't know how to save $field of type ${field.name} to Snowflake")
         }
         val nullable = if (field.nullable) "" else "NOT NULL"
-        sb.append(s""", ${name.replace("\"", "\\\"")} $typ $nullable""".trim)
+        sb.append(s""", ${if (isQuoted(name)) name else quotedName(name)} $typ $nullable""".trim)
```
etduwx: Yes, the escaping of the quotes is necessary for Snowflake to correctly parse the queries. Let's maybe change the above to something like:

`val formattedIdentifierName = if (isQuoted(name) name else quotedName(name)
  sb.append(s""",formattedName.replace(..........),.....)`

etduwx:
@mckim0928  Thanks again for making the changes. See comments inline.

Basically, some small logic fixes, and let's add some actual test cases into the Integration Suite. That way we can be confident that things run well. :)

src/test/scala/net/snowflake/spark/snowflake/SnowflakeSourceSuite.scala
```diff
@@ -321,8 +321,22 @@ class SnowflakeSourceSuite extends BaseTest {
     val createTableCommand =
       DefaultSnowflakeWriter.createTableSql(df, MergedParameters.apply(defaultParams)).trim
     val expectedCreateTableCommand =
-      """CREATE TABLE IF NOT EXISTS test_table (long_str VARCHAR(512),""" +
-        """ short_str VARCHAR(10), default_str STRING)"""
+      """CREATE TABLE IF NOT EXISTS test_table ("long_str" VARCHAR(512),""" +
```
etduwx: Let's add some tests into the IntegrationTests, too, i.e., under `src/it`. :)

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
mckim0928: Totally. Thank you so much for the thorough explanation. The regex pattern now includes lowercase letters as well, which would ensure identifier legality.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
mckim0928: ahh my mistake leaving that out here. fixed!

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
mckim0928: gotcha! Also modified the test case in one of the unit tests, `SnowflakeSourceSuite.scala:324-325` and added in another case that ensures explicitly quoted inputs are not uppercased in `SnowflakeSourceSuit.scala:329`

mckim0928:
@etduwx 

Thanks again for being so responsive and guiding me through the required logic. I believe all of the logic suggested is now implemented with an added unit test case (replied on your previous inline comments). 

Waiting on the SF account credentials stuff and there are chances I might not be able to get to adding / running an IT suit...but I was thinking we could add some test cases under 'SnowflakeIntegrationSuite.scala'? Or would you recommend adding it somewhere else?

etduwx:
@mckim0928 If you don't have credentials access, no worries, I can add a test myself. Thanks for your contribution!

mckim0928:
@etduwx Thank you so much for all your help! Do you think these changes could be merged soon? Please let me know if there needs to be any changes, or if I could help out in any other way.

PR #15, opened 2017-08-30
Azureexternalstage

linchan825:
Support the use of Azure storage as the external stage for data loading/unloading to Snowflake.

### DIFF
--- a/project/SparkSnowflakeBuild.scala
+++ b/project/SparkSnowflakeBuild.scala
@@ -72,6 +72,7 @@ object SparkSnowflakeBuild extends Build {
         "com.amazonaws" % "aws-java-sdk-core" % "1.10.22" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
         "com.amazonaws" % "aws-java-sdk-s3" % "1.10.22" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
         "com.amazonaws" % "aws-java-sdk-sts" % "1.10.22" % "test" exclude("com.fasterxml.jackson.core", "jackson-databind"),
+        "com.microsoft.azure" % "azure-storage" % "5.0.0" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
         // We require spark-avro, but avro-mapred must be provided to match Hadoop version.
         // In most cases, avro-mapred will be provided as part of the Spark assembly JAR.
         // "com.databricks" %% "spark-avro" % "2.0.1",
--- a/src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
@@ -30,16 +30,18 @@ private[snowflake] object CloudCredentialsUtils {
    * Generates a credentials string for use in Snowflake COPY in-out statements.
    */
   def getSnowflakeCredentialsString(sqlContext: SQLContext,
-                                          params: MergedParameters): String = {
-    if (params.rootTempDir.startsWith("file://")) {
-      "-- file URL, no creds needed"
-    } else if (params.rootTempDir.startsWith("wasb://") ||
-               params.rootTempDir.startsWith("wasbs://")) {
-      val creds = getAzureStorageCreds(params)
-      getSnowflakeCredentialsStringForAzure(creds)
-    } else {
-      val creds = getAWSCreds(sqlContext, params)
-      getSnowflakeCredentialsStringForAWS(creds)
+                                    params: MergedParameters): String = {
+    params.rootTempDirStorageType match {
+      case FSType.LocalFile =>
+        "-- file URL, no creds needed"
+      case FSType.Azure =>
+        val creds = getAzureCreds(params)
+        getSnowflakeCredentialsStringForAzure(creds)
+      case FSType.S3 =>
+        val creds = getAWSCreds(sqlContext, params)
+        getSnowflakeCredentialsStringForAWS(creds)
+      case _ =>
+        ""
     }
   }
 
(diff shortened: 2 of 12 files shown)


src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
```diff
@@ -19,26 +19,31 @@ package net.snowflake.spark.snowflake
...
 import org.apache.spark.sql.SQLContext
 
-private[snowflake] object AWSCredentialsUtils {
+private[snowflake] object CloudCredentialsUtils {
   /**
    * Generates a credentials string for use in Snowflake COPY in-out statements.
    */
   def getSnowflakeCredentialsString(sqlContext: SQLContext,
-                                    params: MergedParameters): String = {
+                                          params: MergedParameters): String = {
```
etduwx: Minor nitpick: Would it be possible to preserve original alignment? :)

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
```diff
@@ -19,26 +19,31 @@ package net.snowflake.spark.snowflake
...
     if (params.rootTempDir.startsWith("file://")) {
       "-- file URL, no creds needed"
+    } else if (params.rootTempDir.startsWith("wasb://") ||
```
etduwx: I wonder if we should add a lazy boolean in `MergedParameters` that will signal Azure vs. S3, i.e. something like `isAzure`. This would alleviate the need to do constant string checking. That method can later also be extended to resolve the stage-type for when `tempDir` isn't provided, i.e. when using the internal stage, as well as include other checks.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -97,7 +118,9 @@ object Utils {
   def checkThatBucketHasObjectLifecycleConfiguration(
       tempDir: String,
       s3Client: AmazonS3Client): Unit = {
-    if (tempDir.startsWith("file://")) {
+    if (tempDir.startsWith("file://") ||
```
etduwx: How about doing the a single check for the inverse, i.e. for S3?

etduwx:
Thanks, @linchan825 . Looks good to me, only had a couple of very minor comments. My only worry is that we've created a lot of specialized methods for the different storage providers, i.e. `doXforS3()` and `doXforAzure()`, which makes maintainability a little harder, but that probably makes sense for now since we're likely not going to add a third one anytime soon, and polymorphism would be overkill for now. :)

The good news is that both new failing tests are related only to a JDBC bug, for which I filed a separate JIRA. When you get an account with which to hook up integration tests, please add the necessary changes to IntegrationSuiteBase.scala to permit running the tests on both S3 and Azure, and be sure to also modify the snowflake.conf.example file to reflect the new conf-file variables the tests will look for. We'll merge this once we have the tests working. 

Thanks again!

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
```diff
@@ -19,26 +19,31 @@ package net.snowflake.spark.snowflake
...
+               params.rootTempDir.startsWith("wasbs://")) {
+      val creds = getAzureStorageCreds(params)
```
MarcinZukowski: np: you could use `getAzureCreds` for consistency

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
```diff
@@ -55,17 +60,36 @@ private[snowflake] object AWSCredentialsUtils {
        |""".stripMargin.trim
   }
 
-  def getCreds(sqlContext: SQLContext,
-               params : MergedParameters) : AWSCredentials = {
+  def getSnowflakeCredentialsStringForAzure(
+        creds : StorageCredentialsSharedAccessSignature) : String = {
+    val sasToken = creds.getToken()
```
MarcinZukowski: Just to be safe, I assume Azure tokens use only alphanumeric chars, so no escaping needed?

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -18,6 +18,7 @@
 package net.snowflake.spark.snowflake
 
 import com.amazonaws.auth.{AWSCredentials, BasicSessionCredentials}
+import com.microsoft.azure.storage.StorageCredentialsSharedAccessSignature
```
MarcinZukowski: I have quite a bit of concern here that we're introducing a dependency on Azure here...

project/SparkSnowflakeBuild.scala
```diff
@@ -72,6 +72,7 @@ object SparkSnowflakeBuild extends Build {
         "com.amazonaws" % "aws-java-sdk-core" % "1.10.22" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
         "com.amazonaws" % "aws-java-sdk-s3" % "1.10.22" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
         "com.amazonaws" % "aws-java-sdk-sts" % "1.10.22" % "test" exclude("com.fasterxml.jackson.core", "jackson-databind"),
+        "com.microsoft.azure" % "azure-storage" % "5.0.0" % "provided" exclude("com.fasterxml.jackson.core", "jackson-databind"),
```
MarcinZukowski: Here. In particular, we assume azure-storage is already installed in Spark.

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
linchan825: Changed.

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
linchan825: I think it is cool but in that case I don't want to make it a boolean but a enum. Essentially, we have 3 types of filesystem here: local file, S3 and Azure. Also it will be a bit more extensible later. 
Let me make that change.

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
linchan825: Changed.

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
linchan825: The SAS token will be something like below. AFAIK, the characters doesn't need any escaping. 
?sv=2016-05-31&ss=b&srt=sco&sp=rwdlac&se=2018-09-13T00:27:47Z&st=2017-06-19T16:27:47Z&spr=https,http&sig=aJc9WMofMGM2r90OO8Lw%2BzP%2BWC%2FwxDeP3ESf64JEfro%3D

project/SparkSnowflakeBuild.scala
linchan825: I am assuming this assumption is the same for S3? If users want to use Azure, they need to make sure Azure libraries are included in their classpath, so is S3 right?

PR #19, opened 2018-01-12
Upgrade spark 2.1.0 to 2.2.0

binglihub:


project/SparkSnowflakeBuild.scala
```diff
@@ -45,10 +45,10 @@ object SparkSnowflakeBuild extends Build {
       organization := "net.snowflake",
       scalaVersion := sys.props.getOrElse("SPARK_SCALA_VERSION", default = "2.11.7"),
       crossScalaVersions := Seq("2.10.5", "2.11.7"),
-      sparkVersion := "2.1.0",
+      sparkVersion := "2.2.0",
       testSparkVersion := sys.props.get("spark.testVersion").getOrElse(sparkVersion.value),
-      testHadoopVersion := sys.props.get("hadoop.testVersion").getOrElse("2.2.0"),
-      javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
+      testHadoopVersion := sys.props.get("hadoop.testVersion").getOrElse("2.6.0"),
+      javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
```
etduwx: Let's try to keep Java 1.6 if possible (for backwards compatibility reasons). Our current dependency is Java 6; we had problems with customers even when we tried to increase this to Java 7.

src/it/scala/net/snowflake/spark/snowflake/FilterPushdownIntegrationSuite.scala
```diff
@@ -67,22 +67,22 @@ class FilterPushdownIntegrationSuite extends IntegrationSuiteBase {
...
-         | )
-         | using net.snowflake.spark.snowflake
-         | options(
-         |   $connectorOptionsString
-         |   , dbtable \"$test_table\"
-         | )
-       """.stripMargin
-    )
-  }
+//  override def beforeEach(): Unit = {
```
etduwx: Can delete this block completely if unnecessary.

src/main/java/net/snowflake/spark/snowflake/parquet/FileFromStream.java
```diff
@@ -0,0 +1,31 @@
+package net.snowflake.spark.snowflake.parquet;
```
etduwx: This work should be on a separate branch; we should merge the changes from the parquet feature branch and this version upgrade separately. Can we separate these commits?

src/main/java/net/snowflake/spark/snowflake/parquet/SeekableInputStreamBuilder.java
```diff
@@ -0,0 +1,104 @@
+package net.snowflake.spark.snowflake.parquet;
```
etduwx: Same comment as above; should be separate from the version changes.

project/SparkSnowflakeBuild.scala
ahepburn: really? Java 1.6 reached end of life in 2013... I'd looked into making equivalent changes locally and decided it wasn't worth the effort of PR'ing/etc given the legacy versions of many things...

Per [Spark documentation](https://spark.apache.org/docs/2.2.0/): 

```Note that support for Java 7, Python 2.6 and old Hadoop versions before 2.6.5 were removed as of Spark 2.2.0.```

If you need to maintain support for customers who are relying on legacy versions, then perhaps consider maintaining 2 versions of this connector -- the legacy one, and a version that is current with the ecosystem.

PR #26, opened 2018-03-16
Column mapping

binglihub:
Column mapping with auto pushdown

### DIFF
--- a/project/SparkSnowflakeBuild.scala
+++ b/project/SparkSnowflakeBuild.scala
@@ -83,7 +83,7 @@ object SparkSnowflakeBuild extends Build {
         // },
 
         "com.google.guava" % "guava" % "14.0.1" % "test",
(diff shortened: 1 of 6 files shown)


src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
```diff
@@ -76,6 +82,9 @@ class DefaultSource(jdbcWrapper: JDBCWrapper, s3ClientFactory: AWSCredentials =>
       data: DataFrame): BaseRelation = {
 
     val params = Parameters.mergeParameters(parameters)
+    //check spark version for push down
+    if(parameters.getOrElse("autopushdown","on")=="on")
```
etduwx: use the autoPushdown: Boolean function here

PR #65, opened 2018-08-21
Do not double-quote strings that are already quoted with backticks

bclune:
Currently, the `columnmap` option forces column names to be wrapped
in double-quotes, even when the column name is already quoted with
backticks (e.g. the column name is a reserved word like `values`).

This commit accepts backtick-quoted strings as quoted when calling
`Utils.ensureQuoted()` or `Utils.isQuoted()`.

Note that I have not run the integration tests for my branch, but that
`build/sbt test` is green.


binglihub:
This bug was fixed in the v2.4.5

PR #95, opened 2018-12-16
Upgrade sbt and its plugins

aosagie:
Hey. The older version of sbt isn't working well on my setup so I upgraded everything.

Unfortunately, it doesn't look like sbt-spark-package is going to be updated to support sbt 1.x anytime soon (see: https://github.com/databricks/sbt-spark-package/pull/41). So, I just removed it. Things still work fine on my end without it. Is it necessary?


binglihub:
@aosagie we are working on upgrading sbt also, will update it later.

binglihub:
We upgraded sbt to 1.2.8.

PR #108, opened 2019-03-19
Fix exception when dealing with java.sql.Time

aosagie:
Hi,

My org frequently queries data that's in the TIME format. Right now, it causes jobs to crash:
Screen Shot 2019-03-19 at 5 37 08 PM

This PR:
1) Casts java.sql.Time to String (preventing the crash from happening)
2) Cleans up build.sbt with modern sbt syntax

We could have chosen to convert java.sql.Time to Spark's TimestampType. However, this would cause a lot of confusion among analysts who would attempt to compare "real" timestamps to "java.sql.Time" timestamps (in which the date portion is zero'd out). That implicit conversion isn't intuitive so we're better off keeping it as a string and allowing analysts to explicitly convert.

Thanks!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,10 +14,7 @@
  * limitations under the License.
  */
 
-import sbt._
-import Keys._
 import sbt.Tests.{Group, SubProcess}
-import sbtrelease.ReleasePlugin.autoImport._
 
 import scala.util.Properties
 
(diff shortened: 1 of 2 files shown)


aosagie:
Hi @binglihub 
Is there anything I can do to help move this along? Issues or formatting to fix? The error's still hitting us in production.

PR #149, opened 2019-09-24
Support pushdown for decimal binary arithmetic operators

Mingli-Rui:


src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/NumericStatement.scala
```diff
@@ -28,7 +28,16 @@ private[querygeneration] object NumericStatement {
...
+        case PromotePrecision(child) => convertStatement(child, fields)
+
+        case CheckOverflow(child, t) =>
+          MiscStatement.getCastType(t) match {
+            case None =>
+              convertStatement(child, fields)
+            case Some(cast) =>
+              ConstantString("CAST") +
+                blockStatement(convertStatement(child, fields) + "AS" + cast)
+          }
```
binglihub: It is fine. Don't need change it this time.

Normally, we don't explicitly write None in the case match statement. It can be
```scala
MiscStatement.getCastType(t) match {
   case Some(cast) => ConstantString("CAST") + blockStatement(convertStatement(child, fields) + "AS" + cast)
   case _ => convertStatement(child, fields)
}
```

Mingli-Rui: Got it. I will update it in next fix.

PR #152, opened 2019-10-02
Enhance pushdown

Mingli-Rui:


Mingli-Rui:
two commits are there, close and re-submit.

PR #154, opened 2019-10-03
Support to retrieve data with SELECT instead of COPY UNLOAD

Mingli-Rui:


src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -0,0 +1,402 @@
...
+    thisConnectorOptionsNoTable += ("query_result_format" -> "arrow")
+    thisConnectorOptionsNoTable += ("time_output_format" -> "HH24:MI:SS.FF")
+
+    setupNumberTable()
+    setupStringBinaryTable()
+    setupDateTimeTable()
+    setupTimestampTable()
+    setupLargeResultTable()
+
+    SnowflakeConnectorUtils.enablePushdownSession(sparkSession)
```
etduwx: I think we don't need to turn on pushdown for this test suite, and we don't really care about the query generated.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -0,0 +1,402 @@
...
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("dbtable", s"$test_table_timestamp")
+      .load()
+
+    tmpdf.createOrReplaceTempView("test_table_timestamp")
+  }
+
+  private val largeStringValue =
+    s"""spark_connecor_test_large_result_1234567890
```
etduwx: typo: "connector"

src/it/scala/net/snowflake/spark/snowflake/VariantTypeSuite.scala
```diff
@@ -4,6 +4,8 @@ import org.apache.spark.sql.types._
 import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
 import org.apache.spark.sql.{Row, SaveMode}
 
+import scala.util.parsing.json.JSON
```
etduwx: Can we move this to ResultSetSuite?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -177,6 +179,21 @@ private[snowflake] class JDBCWrapper {
     DriverManager.getConnection(jdbcURL, jdbcProperties)
   }
 
+  private def setupJDBCOptions(params: MergedParameters, jdbcProperties: Properties): Unit = {
```
etduwx: Why are only these two options in this method? If we define a method, we should set all of the properties in here.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -140,19 +143,46 @@ private[snowflake] case class SnowflakeRelation(
...
                                  ): RDD[T] = {
-    val format: SupportedFormat =
-      if (Utils.containVariant(resultSchema)) SupportedFormat.JSON
-      else SupportedFormat.CSV
+    if (params.useCopyUnload) {
+      val format: SupportedFormat =
+        if (Utils.containVariant(resultSchema)) SupportedFormat.JSON
+        else SupportedFormat.CSV
+
+      val rdd: RDD[String] = io.readRDD(sqlContext, params, statement, format)
```
etduwx: nit: I know this is not part of the PR, but can we refactor io -> IO?

src/it/scala/net/snowflake/spark/snowflake/VariantTypeSuite.scala
Mingli-Rui: Do you mean to merge this test case to new test file: SnowflakeResultSetRDDSuite.scala? I think it may better to keep the test case to be separate. According test case name, it mainly test VARIANT type. It is very clear. SnowflakeResultSetRDDSuite is targeted to test new cases. This test case is updated because the data return format is some different but logically correct in the new data reading mode.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
Mingli-Rui: The reason is only these two options are used in this case. It looks that it is not necessary to use a separate function. I will change to set parameter in getConnector().

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
Mingli-Rui: I will file a separate Jira for the code refactor.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
Mingli-Rui: All the query is SELECT * FROM <TABLE>. It will be deleted.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
Mingli-Rui: thanks. correct it

src/it/scala/net/snowflake/spark/snowflake/VariantTypeSuite.scala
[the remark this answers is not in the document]
Mingli-Rui: I have noticed this API is deprecated. I will update it to use ObjectMapper.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -549,6 +556,17 @@ object Parameters {
 
     def bindVariableEnabled: Boolean = isTrue(parameters(PARAM_BIND_VARIABLE))
 
+    def useCopyUnload: Boolean = isTrue(parameters(PARAM_USE_COPY_UNLOAD))
+
+    def expectedPartitionSize: Long = {
+      try {
+        (parameters(PARAM_EXPECTED_PARTITION_SIZE_IN_MB).toDouble * 1024 * 1024).toLong
```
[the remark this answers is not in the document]
Mingli-Rui: Lower boundary is no problem. No matter how small it is, at least one file chunk is included. If user configures a super large value. One worker node will handle all the data.From functionality perspective, it is no problem. If we want to add an upper boundary, the problem is what's it should be?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -177,6 +179,21 @@ private[snowflake] class JDBCWrapper {
...
+    if (params.parameters.get("query_result_format").isDefined) {
+      jdbcProperties.put("query_result_format",
+        params.parameters.get("query_result_format").get)
+    }
+
+    if (params.parameters.get("time_output_format").isDefined) {
+      jdbcProperties.put("time_output_format",
+        params.parameters.get("time_output_format").get)
+    }
```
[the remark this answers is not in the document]
Mingli-Rui: I understand your idea, it's doable. But I want to play it simply to keep current change.
Look into the function of getConnector(). For some of the parameters, JDBC parameter names are used directly, but for some of them, separate fields in parameter are used. Even for fields in parameter, both "match" and "if" can be seen.
My understanding is that, if the parameter name in Spark Connector is different to the name in JDBC, a separate field name can be used e.g. "sfrole" -> "role", otherwise, it may not be necessary.
In the future, if we want to enable spark connector user can set any JDBC properties, it may make the parameter class very big.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -140,19 +143,46 @@ private[snowflake] case class SnowflakeRelation(
...
+          log.debug("partition  " + index + ": " + resultSet.toString())
+        }
+      }
+
+      new SnowflakeResultSetRDD[T](
+        resultSchema,
+        sqlContext.sparkContext,
+        resultSetSerializables
+      )
+    }
```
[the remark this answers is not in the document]
Mingli-Rui: Good idea. Follow it.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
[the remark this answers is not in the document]
binglihub: In this case, we need document it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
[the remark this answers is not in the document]
binglihub: That is legacy issues. At least, don't use String value directly. Try to reference this values from one constant.
We have no plan to directly forward JDBC parameters because JDBC should be transparent in Spark Connector. If you directly forward parameters, customer can crash Spark Connector.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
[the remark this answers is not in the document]
Mingli-Rui: agree

PR #161, opened 2019-11-12
Support column name with (.)dot for "column_mapping = name"

Mingli-Rui:


binglihub:
@Mingli-Rui are you still working on this PR? Can we close it?

PR #163, opened 2019-12-09
Support supper large data

Mingli-Rui:
For details description refer to Jira SNOW-119245

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
@@ -17,6 +17,13 @@
 
 package net.snowflake.spark.snowflake
 
+import java.io.{ByteArrayInputStream, InputStream}
+import java.util.Base64
+
+import net.snowflake.client.core.QueryResultFormat
+import net.snowflake.client.jdbc.internal.apache.arrow.memory.RootAllocator
+import net.snowflake.client.jdbc.internal.apache.arrow.vector.VectorSchemaRoot
+import net.snowflake.client.jdbc.internal.apache.arrow.vector.ipc.ArrowStreamReader
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types._
@@ -26,9 +33,11 @@ import net.snowflake.spark.snowflake.Parameters.MergedParameters
 import net.snowflake.spark.snowflake.io.SupportedFormat.SupportedFormat
 import net.snowflake.spark.snowflake.io.SupportedFormat
 import net.snowflake.spark.snowflake.DefaultJDBCWrapper.DataBaseOperations
+
 import scala.reflect.ClassTag
-import net.snowflake.client.jdbc.{SnowflakeResultSet, SnowflakeResultSetSerializable}
+import net.snowflake.client.jdbc.{SnowflakeResultSet, SnowflakeResultSetSerializable, SnowflakeResultSetSerializableV1}
 import net.snowflake.spark.snowflake.io.SnowflakeResultSetRDD
+
 import scala.collection.JavaConversions
 
 /** Data Source API implementation for Amazon Snowflake database tables */
@@ -186,12 +195,7 @@ private[snowflake] case class SnowflakeRelation(
     val resultSetSerializables: Array[SnowflakeResultSetSerializable] =
       JavaConversions.asScalaBuffer(resultSet.asInstanceOf[SnowflakeResultSet]
         .getResultSetSerializables(params.expectedPartitionSize)).toArray
-
-    resultSetSerializables.zipWithIndex.map {
-      case (resultSet, index) => {
-        log.debug("partition  " + index + ": " + resultSet.toString())
-      }
-    }
+    printStatForSnowflakeResultSetRDD(resultSetSerializables)
 
     new SnowflakeResultSetRDD[T](
       resultSchema,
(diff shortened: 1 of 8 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -201,6 +201,112 @@ private[snowflake] case class SnowflakeRelation(
...
 
+  // Print result set statistic information
+  private def printStatForSnowflakeResultSetRDD(
+     resultSetSerializables: Array[SnowflakeResultSetSerializable]) : Unit = {
+    var totalRowCount : Long = 0
+    var totalCompressedSize : Long = 0
+    var totalUnCompressedSize : Long = 0
+    var totalURLCount : Long = 0
+
+    resultSetSerializables.zipWithIndex.map {
```
binglihub: It seems like you don't need index from here. and if you only want to scan the collection without return value, you can use `foreach` instead of `map`

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -186,12 +191,7 @@ private[snowflake] case class SnowflakeRelation(
     val resultSetSerializables: Array[SnowflakeResultSetSerializable] =
       JavaConversions.asScalaBuffer(resultSet.asInstanceOf[SnowflakeResultSet]
         .getResultSetSerializables(params.expectedPartitionSize)).toArray
-
-    resultSetSerializables.zipWithIndex.map {
-      case (resultSet, index) => {
-        log.debug("partition  " + index + ": " + resultSet.toString())
-      }
-    }
+    printStatForSnowflakeResultSetRDD(resultSetSerializables)
```
binglihub: Is it going to download result to master? or just meta

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
Mingli-Rui: I will update it to use FOREACH.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
Mingli-Rui: It's only metadata. This function only collects statistic from the metadata and print it in log.

src/main/scala/net/snowflake/spark/snowflake/io/SnowflakeRDD.scala
```diff
@@ -11,22 +11,39 @@ class SnowflakeRDD(
...
+                    downloadFile: (String) => InputStream,
+                    expectedPartitionCount: Int
                   ) extends RDD[String](sc, Nil) {
 
-  @transient private val FILES_PER_PARTITION = 2
+  @transient private val MIN_FILES_PER_PARTITION = 2
 
   override def compute(split: Partition, context: TaskContext): Iterator[String] = {
     val stringIterator = new SFRecordReader(format)
+    stringIterator.setDownloadFunction(downloadFile)
```
etduwx: Seems like these functions can be part of the constructor.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -52,7 +54,8 @@ import scala.collection.JavaConversions._
 
 object CloudStorageOperations {
   private[io] final val DEFAULT_PARALLELISM = 10
-  private[io] final val S3_MAX_RETRIES = 3
+  private[io] final val S3_MAX_RETRIES = 10
```
etduwx: 10 retries, with the sleep time doubled every time. That means we can end up sleeping for 2000 seconds in the last retry. That means if anyone has this mode turned on and they have a more simple error, they can wait a long time before they find out. Should this be the default value? Also, should this be configurable?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -201,6 +204,112 @@ private[snowflake] case class SnowflakeRelation(
...
+        val root = reader.getVectorSchemaRoot
+        root.getFieldVectors
+        while (reader.loadNextBatch()) {
+          rowCount += root.getRowCount
+          root.clear
+        }
+
+        root.clear
+        reader.close
+        is.close
```
etduwx: These should be in a finally clause?

src/main/scala/net/snowflake/spark/snowflake/io/SnowflakeRDD.scala
Mingli-Rui: My initial idea is to integrate the download function into the constructor. But, it is some difficult, so I separate it from constructor. There are another two use cases that the InputStream is for some local files not from network. In this way, the code change is clearer, and lower risk. Currently, SFRecordReader can handle both Filename and InputStream.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
Mingli-Rui: The initial backoff is 100ms not 1s, so I think the last sleep should be 2**9 * 0.1 = 50s. Currently, it is not configurable. The reason is that we have introduced retry on spark connector which is configurable. It sounds no necessary to introduce two max retry count.  I admit it should be better to use smaller API retry count. I will change the API max retry count as 6. 2**6 * 0.1 = 6.4s. What do you think.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
Mingli-Rui: You are right! I will move them to finally clause.

PR #167, opened 2019-12-18
enable column-mapping by name in streaming sink

binglihub:


binglihub:
@Mingli-Rui please review it, to make sure there is no impact on your work.

src/it/scala/net/snowflake/spark/snowflake/streaming/StreamingSuite.scala
```diff
@@ -130,7 +130,7 @@ class StreamingSuite extends IntegrationSuiteBase {
 
   }
 
-  test("Test streaming writer") {
```
binglihub: I disable all streaming tests by default, only run it manually

PR #175, opened 2020-01-16
Use const string as spark connector short name.

Mingli-Rui:
SNOW-132989

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
@@ -49,7 +49,12 @@ object Utils {
     */
   val SNOWFLAKE_SOURCE_NAME = "net.snowflake.spark.snowflake"
 
-  val VERSION = "2.5.6"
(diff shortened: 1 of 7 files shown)


src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -49,7 +49,12 @@ object Utils {
     */
   val SNOWFLAKE_SOURCE_NAME = "net.snowflake.spark.snowflake"
 
-  val VERSION = "2.5.6"
+  /**
+    * Short literal name of SNOWFLAKE_SOURCE_NAME
+    */
+  val SNOWFLAKE_SOURCE_SHORT_NAME = "snowflake"
+
+  val VERSION = "2.5.8"
```
binglihub: Did you change this version name in last release?

Mingli-Rui: No. It was 2.5.6

PR #176, opened 2020-01-24
Support pushdown for OUTER JOIN

Mingli-Rui:
Enhance below 2 cases:
1. Support FULL OUTER JOIN pushdown.
2. Support LEFT/RIGHT JOIN even if there are not-null columns.

SNOW-82759


etduwx:
Approved, even though it has already been merged. :)

PR #178, opened 2020-02-05
Fix travis script, update JDBC to 3.12.0, introduce max file per partition

Mingli-Rui:
SNOW-136399
SNOW-136396

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ matrix:
   include:
   - jdk: openjdk8
     scala: 2.12.8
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -7,10 +7,10 @@ matrix:
   include:
   - jdk: openjdk8
     scala: 2.12.8
-    env: SPARK_VERSION="2.4.0" INTEGRATION_TESTS="true" "SNOWFLAKE_TEST_ACCOUNT"="aws" "SPARK_CONN_ENV_USE_COPY_UNLOAD"="false"
+    env: SPARK_VERSION="2.4.0" INTEGRATION_TESTS=true SNOWFLAKE_TEST_ACCOUNT=aws SPARK_CONN_ENV_USE_COPY_UNLOAD=false
```
etduwx: Why do we not use the quotes anymore?

Mingli-Rui: It looks this will cause the travis build failure. We didn't see the failure before.

PR #179, opened 2020-02-06
Enable pushdown for date_trunc

jjchacona:
Support for pushdown date_trunc.

Snowflake: 
https://docs.snowflake.net/manuals/sql-reference/functions/date_trunc.html

Spark 2.4: http://spark.apache.org/docs/latest/api/python/_modules/pyspark/sql/functions.html#date_trunc


Mingli-Rui:
Can you create a pull request with only one COMMIT?

Mingli-Rui:
In addition, please only add necessary change only.

PR #185, opened 2020-02-12
Support to read Double INF/-INF/NaN values.

Mingli-Rui:
SNOW-137832

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
   .settings(
     name := "spark-snowflake",
     organization := "net.snowflake",
-    version := "2.5.9-spark_2.4",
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -29,7 +29,7 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
   .settings(
     name := "spark-snowflake",
     organization := "net.snowflake",
-    version := "2.5.9-spark_2.4",
+    version := "2.6.0-spark_2.4",
```
etduwx: Is this enabling Arrow by default? Or should we use 2.5.10 if we're not?

Mingli-Rui: This is only version change. It doesn't enable arrow. To enable arrow format, what we should do is to change the default value of 'use_copy_unload' as false.

PR #190, opened 2020-03-03
Support dot(.) in column name

Mingli-Rui:
In Spark, if there are dot(.) in the column name, the column name should be enclosed with backtick(`). Spark connector follow this rule. 
FYI.
https://forums.databricks.com/questions/30739/how-to-deal-with-column-name-with-dot-in-pyspark-d.html
SNOW-116375

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeWriter.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeWriter.scala
@@ -108,7 +108,13 @@ private[snowflake] class SnowflakeWriter(jdbcWrapper: JDBCWrapper) {
                                    params: MergedParameters): DataFrame =
     params.columnMap match {
       case Some(map) =>
-        val names = map.keys.toSeq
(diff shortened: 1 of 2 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeWriter.scala
```diff
@@ -108,7 +108,13 @@ private[snowflake] class SnowflakeWriter(jdbcWrapper: JDBCWrapper) {
                                    params: MergedParameters): DataFrame =
     params.columnMap match {
       case Some(map) =>
-        val names = map.keys.toSeq
+        // Enclose column name with backtick(`) if dot(.) exists in column name
+        val names = map.keys.toSeq.map(name =>
+          if (name.contains(".")) {
+            s"`$name`"
```
binglihub: make sure there is no "`" in the column name of Snowflake tables.

Mingli-Rui: confirmed. There is no "`" in the snowflake table column name.

PR #195, opened 2020-03-11
Support Snowflake GCP account

Mingli-Rui:
All functionalities are supported except Spark streaming and external stage


build.sbt
```diff
@@ -18,6 +18,7 @@ import scala.util.Properties
 
 val sparkVersion = "2.4.0"
 val testSparkVersion = sys.props.get("spark.testVersion").getOrElse(sparkVersion)
+val sparkConnectorVersion = "2.6.1"
```
binglihub: I think supporting GCP is a major feature, the version can be 2.7.0

Mingli-Rui: agree.

PR #197, opened 2020-03-16
SNOW-124751

binglihub:
SNOW-124751 enable column-mapping by name in streaming sink


Mingli-Rui:
duplicate to https://github.com/snowflakedb/spark-snowflake/pull/167/commits

PR #200, opened 2020-03-20
Rename SnowflakeSink.schema as SnowflakeSink.cachedSchema

Mingli-Rui:
SNOW-146866
In Spark 3.0, the parent class of SnowflakeSink has a conflict name 'schema'.


Mingli-Rui:
It is for Spark 3.0 support. In spark 3.0, the parent class of SnowflakeSink has a function named schema. There is name conflict.

PR #201, opened 2020-03-23
Spark 3.0 support

Mingli-Rui:
https://snowflakecomputing.atlassian.net/browse/SNOW-88836

Core changes:
1. Support LIKE clause with ESCAPE character, for example,
   LIKE “100^%” ESCAPE '^'. New test cases are added to test it.
2. The interface function for LogicalPlan.argString() is changed to
   argString(maxFields: Int) which is used in SnowflakeTelemetry to
   convert the logical plan to telemetry message. From Spark 3.0,
   at most 1024 nodes are included in one telemetry message.
3. New parameter “JoinHint“ is introduced for the constructor of Join.
   Snowflake doesn’t support JoinHint, so it is ignored.
4. New parameter “nullOnOverflow“ is introduced for the constructor of
   CheckOverflow. CheckOverflow is pushdown as CAST in snowflake.
   Snowflake doesn’t support this parameter, so it is ignored.


Mingli-Rui:
This PR is the essential change for Spark 3.0. It is NOT merged to current master branch. I create a new branch 'master_future' to preserve this change and code review. We'll release this change when Apache 3.0 GA is released. If some customer wants to try Spark 3.0, we can instruct them to build with branch 'master_future' or 'spark_3.0'.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/NumericStatement.scala
```diff
@@ -75,7 +75,7 @@ private[querygeneration] object NumericStatement {
 
       case PromotePrecision(child) => convertStatement(child, fields)
 
-      case CheckOverflow(child, t) =>
+      case CheckOverflow(child, t, _) =>
```
binglihub: We may need handle this new argument.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -15,6 +15,7 @@ import net.snowflake.spark.snowflake.TelemetryTypes.TelemetryTypes
 object SnowflakeTelemetry {
 
   private val TELEMETRY_SOURCE = "spark_connector"
+  private val MAX_OUTPUT_NODE_COUNT = 1024
```
etduwx: Just for my understanding, what is this?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -94,7 +95,8 @@ object SnowflakeTelemetry {
       case Project(fields, _) =>
         argsNode.set("fields", expressionsToJson(fields))
 
-      case Join(_, _, joinType, Some(condition)) =>
+      // Snownflake doesn't support HINT for join, so just ignore the hint.
```
etduwx: nit: typo

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/NumericStatement.scala
Mingli-Rui: Because snowflake doesn't support this option, what we can do is disable push-down for this case. But this is current behavior, so I prefer to keep it. In addition, I will file a Jira to track it, if snowflake supports this feature, we can support it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
Mingli-Rui: Telemetry service may send the logical plan to snowflake. And LogicalPlan.argString() is called to convert the plan as a String. From spark 3.0, this function needs a parameter: maxFields which can limit the fields count to be converted to a string. So from our side, we just provide a magic number like 1024 at the max count.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
Mingli-Rui: correct it.

PR #203, opened 2020-03-24
Github test platform POC

Mingli-Rui:


.github/workflows/IntegrationTest.yml
```diff
@@ -0,0 +1,27 @@
...
+        java-version: 1.8
+    - name: Decrypt snowflake.json for testing
+      run: ./.github/scripts/decrypt_secret.sh
+      env:
+        SNOWFLAKE_TEST_PROFILE_SECRET: ${{ secrets.SNOWFLAKE_TEST_PROFILE_SECRET }}
+    - name: Run tests
+      run: |
+          sbt compile
+          sbt test
+          sbt it:test
```
binglihub: _sbt it: test_ includes _sbt compile_ and _sbt test_ . don't need call them two times

.github/docker/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+      - SNOWFLAKE_TEST_CONFIG=/Users/ec2-user/snowflake.travis.json
+      - SPARK_CONN_ENV_SFCOMPRESS=on
+      - SPARK_CONN_ENV_SFSSL=on
+      - SPARK_CONN_ENV_RUNOPTION=snowflake-stage
+      - SPARK_CONN_ENV_SFDATABASE=testdb_spark
+      - SPARK_CONN_ENV_SFSCHEMA=spark_test
+      - SPARK_CONN_ENV_SFWAREHOUSE=testwh_spark
+      - SPARK_CONN_ENV_SFUSER=test_spark
+      - SPARK_CONN_ENV_PEM_PRIVATE_KEY=$SPARK_CONN_ENV_PEM_PRIVATE_KEY
+      - SPARK_CONN_ENV_SFURL=sfctest0.snowflakecomputing.com
```
binglihub: I think we need hide all login info not only the private key.

Mingli-Rui: got it. I just want to make it work first. long team, I want to pass the encrypted json file and decrypt it in side the container.

PR #205, opened 2020-04-07
Dockerize a spark cluster for GitHub Action test.

Mingli-Rui:


Mingli-Rui:
This Jira dockerizes a spark standalone cluster with docker-compose in github action.
And it can run a test case with a compiled jar file. Next step is to develop the test case and integrate the test compilation in the auto test.
https://snowflakecomputing.atlassian.net/browse/SNOW-151197

binglihub:
what is the test case content?

Mingli-Rui:
Currently, there is only one test case which reads some rows from a snowflake table and write the data to another snowflake table. Next Jira will merge the test case code and integrate them in the test.

PR #206, opened 2020-04-09
Add basic test cases and integrate them in ClusterTest

Mingli-Rui:


Mingli-Rui:
In previous PR, the docker container used a compiled test jar file for test. In this PR, one basic Scalar and one basic python test case are added. They will be compiled, packaged and setup into the docker container in automatically. 
Next step is to enhance the test cases. So, please reviewers not to spend much time on inside of below two test cases files:
ClusterTest/src/main/python/unittest.py
ClusterTest/src/main/scala/net/snowflake/spark/snowflake/ClusterTest.scala

Mingli-Rui:
https://snowflakecomputing.atlassian.net/browse/SNOW-151598

PR #208, opened 2020-04-17
Implement retry when uploading data to cloud storage.

sfc-gh-mrui:


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-153132

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -469,9 +469,21 @@ private[io] object StorageInfo {
...
-  protected val RETRY_SLEEP_TIME_UNIT_IN_MS: Long = 200
+  protected val RETRY_SLEEP_TIME_UNIT_IN_MS: Long = 100
   private var processedFileCount = 0
   protected val connection: Connection
+  protected val maxRetryCount: Int
+  protected val proxyInfo: Option[ProxyInfo]
+
+  protected def retrySleepTimeInMS(retry: Long): Long = {
+    var expectedTime: Long =
+      RETRY_SLEEP_TIME_UNIT_IN_MS * Math.pow(2, retry).toLong
```
sfc-gh-bli, line 480: first time only sleep 100 * 2 ms?  0.2 sec?

sfc-gh-mrui: Correct. The sleep time is exponential increased from 200ms. The max sleep time is 1 minute. “max_retry_count” is configurable.The default value of max_retry_count is 10. The sleep time (ignore noise) will be like: 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 25.6, 51.2, 60. By default, the total sleep time be about 160 seconds. In addition, we add some random noise for the sleeping time. the noise time is at most 20% for the calculated sleep time.

PR #210, opened 2020-04-20
Log request when transferring data for Azure and move to version 2.7.1

sfc-gh-mrui:


sfc-gh-mrui:
To reviewers: focus review the code change in CloudStorageOperations.scala. The other code is for test framework for new version 2.7.1. Core change in CloudStorageOperations.scala: 
 1. log request id for azure cloud.
2. Change the backoff sleep time strategy like XP.

https://snowflakecomputing.atlassian.net/browse/SNOW-153811

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -895,8 +904,37 @@ case class InternalAzureStorage(param: MergedParameters,
             storageInfo(StorageInfo.SMK_ID)
           )
       blob.setMetadata(meta)
-      new CipherOutputStream(blob.openOutputStream(), cipher)
+
```
sfc-gh-ema, line 907: Can you also log how many times we've retried?

sfc-gh-mrui: The retry count has been logged by the caller of this function: createDownloadStreamWithRetry() and uploadPartition() when retry is triggered.

PR #214, opened 2020-05-01
Close connection in Utils.runQuery() and refactor some tests

sfc-gh-mrui:


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-156067

sfc-gh-mrui:
> One query one connection? Please double check runQuery command hasn't be used by other Connector component, no internal usage.

It is not used by production code. There are two usages in integration test.

PR #215, opened 2020-05-04
Support new result format for COPY INTO LOCATION

sfc-gh-mrui:
1. The result is checked to send the telemetry message.
   If the result format is not recognized, log a warning message
2. Send the similar telemetry message for Arrow Format
3. Replace all usages of assert() in production with readable Exception

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -445,7 +445,13 @@ object Parameters {
       * URL pointing to the snowflake database, simply
       *   host:port
       */
-    def sfURL: String = parameters(PARAM_SF_URL)
+    def sfURL: String = {
+      var url = parameters(PARAM_SF_URL)
+      if (url.matches("https?://.*")) {
+        url = url.substring(url.indexOf("//") + 2)
+      }
+      url
+    }
 
     /**
       * Snowflake database name
@@ -549,10 +555,9 @@ object Parameters {
       */
     def setColumnMap(fromSchema: Option[StructType],
                      toSchema: Option[StructType]): Unit = {
-      assert(
-        parameters.get(PARAM_COLUMN_MAP).isEmpty,
-        "Column map is already declared"
-      )
+      if (parameters.get(PARAM_COLUMN_MAP).isDefined) {
+        throw new Exception("Column map is already declared")
+      }
       generatedColumnMap =
         if (columnMapping == "name" && fromSchema.isDefined && toSchema.isDefined) {
           val map = Utils.generateColumnMap(
--- a/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
@@ -223,9 +223,11 @@ object Utils {
   private def mapFromSource(src: Source): Map[String, String] = {
     var map = new mutable.HashMap[String, String]
     for (line <- src.getLines()) {
-
       val index = line.indexOf('=')
-      assert(index > 0, "Can't parse this line: " + line)
+      if (index < 1) {
+        throw new Exception(
+          s"Can't parse the line of $line. The index of '=' is $index")
+      }
       val key = line.substring(0, index).trim.toLowerCase
       val value = line.substring(index + 1).trim
       if (!key.startsWith("#")) {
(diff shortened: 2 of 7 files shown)


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-155599
https://snowflakecomputing.atlassian.net/browse/SNOW-156953

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -445,7 +445,13 @@ object Parameters {
       * URL pointing to the snowflake database, simply
       *   host:port
       */
-    def sfURL: String = parameters(PARAM_SF_URL)
+    def sfURL: String = {
+      var url = parameters(PARAM_SF_URL)
+      if (url.matches("https?://.*")) {
+        url = url.substring(url.indexOf("//") + 2)
+      }
+      url
```
sfc-gh-bli, line 453: You can use Regex Group function
https://www.tutorialspoint.com/javaregex/javaregex_capturing_groups.htm

And do you think we can verify the URL format at here?

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -549,10 +555,9 @@ object Parameters {
       */
     def setColumnMap(fromSchema: Option[StructType],
                      toSchema: Option[StructType]): Unit = {
-      assert(
-        parameters.get(PARAM_COLUMN_MAP).isEmpty,
-        "Column map is already declared"
-      )
+      if (parameters.get(PARAM_COLUMN_MAP).isDefined) {
+        throw new Exception("Column map is already declared")
+      }
```
sfc-gh-bli, line 560: What is the difference between Assert and Exception? Do you want to break the process or not?

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -223,9 +223,11 @@ object Utils {
   private def mapFromSource(src: Source): Map[String, String] = {
     var map = new mutable.HashMap[String, String]
     for (line <- src.getLines()) {
-
       val index = line.indexOf('=')
-      assert(index > 0, "Can't parse this line: " + line)
+      if (index < 1) {
+        throw new Exception(
+          s"Can't parse the line of $line. The index of '=' is $index")
+      }
```
sfc-gh-bli, line 230: Same here

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
```diff
@@ -14,7 +14,7 @@ import org.slf4j.{Logger, LoggerFactory}
 
 import scala.util.Random
 
-private[io] object StageReader {
+private[snowflake] object StageReader {
```
sfc-gh-bli, line 17: Why changed to snowflake?

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
```diff
@@ -122,9 +136,6 @@ private[io] object StageReader {
       System.currentTimeMillis()
     )
     SnowflakeTelemetry.send(conn.getTelemetry)
```
sfc-gh-bli, line 138: Where is the telemetry Data?
SnowflakeTelemetry.send() function only flush the buffer and sends cache the metrics to Snowflake.
To add new telemetry message, you should call SnowflakeTelemetry.addLog()

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
sfc-gh-mrui: It is necessary to make sendEgressUsage() to be called for arrow format.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: I will update to use Regex Group in PR. It is not necessary to verify the URL format except there is a strong reason.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: IMO, Assert should not be used in production code.

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
sfc-gh-mrui: This is not a new telemetry. It's to send the EGRESS for Arrow format.

PR #221, opened 2020-05-28
SNOW-163419 Move Spark Connector release deployment script

sfc-gh-mrui:
Move the script from snowflake repro to Spark Connector repro

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,105 @@
+#!/bin/bash -e
+#
+# Push the Spark Connector to the public maven repository
+# This script needs to be executed by snowflake jenkins job
+#
+
+if [ -z "$GPG_KEY_ID" ]; then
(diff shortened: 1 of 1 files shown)


sfc-gh-ema:
I don't understand, why are we moving this file into spark-snowflake? It doesn't seem like the correct repo for it.

sfc-gh-mrui:
> I don't understand, why are we moving this file into spark-snowflake? It doesn't seem like the correct repo for it.

The reason is from Whitesource project. WhiteSource project requests us to add Whitesource scan to the release build script in jenkins job. To test whether the whitesource script work or not, I have to change the build script multiple time. It's very inconvenient.

All the other components have moved out the deploy script to their own directory or repository. Spark connector is the only component to do actual deployment in Tests/jenkins/MavenPushSparkConnector/run.sh.

So, we'd better to move it to our repository.

PR #224, opened 2020-06-10
SNOW-166185 Send telemetry message if pushdown query fail.

sfc-gh-mrui:


sfc-gh-mrui:
To reviewers: Refer to below Jira for the details PR description
https://snowflakecomputing.atlassian.net/browse/SNOW-166185

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -995,6 +995,7 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
       )""".stripMargin
     )
     awsConn.createStatement.execute("GET @test_move_data/ file:///tmp/test_move_data")
+    awsConn.close()
```
sfc-gh-ema, line 998: I assume this change (and the other changes like this one) are not related to telemetry?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/UnsupportedStatement.scala
```diff
@@ -0,0 +1,49 @@
...
+    throw new SnowflakePushdownUnsupportedException(
+      EXCEPTION_MESSAGE,
+      expr.prettyName,
+      expr.sql,
+      isKnownUnsupportedOperation(expr))
+  }
+
+  // Determine whether the unsupported operation is known or not.
+  private def isKnownUnsupportedOperation(expr: Expression): Boolean = {
+    // The pushdown for UDF is known unsupported
```
sfc-gh-ema, line 44: I'm actually not sure if we want to filter these out. Maybe we can flag them, but these are useful also to capture for statistics. For example, we might want to know how many customers are trying to pushdown UDFs.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/UnsupportedStatement.scala
```diff
@@ -0,0 +1,49 @@
...
+}
+import org.apache.spark.sql.execution.aggregate.ScalaUDAF
+
+/**
+  * This class is used to catch unsupported statement and raise an exception
+  * to stop the push-down to snowflake.
+  */
+private[querygeneration] object UnsupportedStatement {
+
+  final val EXCEPTION_MESSAGE = "Meet unsupported statement"
```
sfc-gh-ema, line 18: Should this actually be an exception? It should be enough to simply send a telemetry message. Exceptions can be big and scary, and we don't want to alarm users if we simply failed to translate something.

sfc-gh-ema: Ok nevermind, I just saw your catch clause up there. This should be fine.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -91,6 +91,40 @@ object SnowflakeTelemetry {
...
+    *
+    * @param plan The logical plan to include the unsupported operations
+    * @param exception The pushdown unsupported exception
+    */
+  def addPushdownFailMessage(plan: LogicalPlan,
+                             exception: SnowflakePushdownUnsupportedException)
+  : Unit = {
+    logger.info(
+      s"""Pushdown fails because of operation: ${exception.unsupportedOperation}
+         | message: ${exception.getMessage}
```
sfc-gh-ema, line 106: You hardcoded this value, so it's always the same, right?

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
sfc-gh-mrui: Correct!
I browsed all connections built in spark connector, I found some connections are not closed yet. So, I close them.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/UnsupportedStatement.scala
sfc-gh-ema: And also nit: change the message to 'pushdown failed'

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: I didn't get your question. what does "You hardcoded this value" refer to?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/UnsupportedStatement.scala
sfc-gh-mrui: 1. It's necessary to raise an exception here and let outer function to catch the exception and send the telemetry message. The reason is that we only need to send telemetry message if there are any snowflake tables. We don't know whether there are any snowflake table here.
2. I will change the message as 'pushdown failed'.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/UnsupportedStatement.scala
sfc-gh-mrui: This is optional. I decided to filter them out because our purpose is to catch unsupported statement and support them in the future. The message for known unsupported statement sounds noise for this purpose.
If we need to know how many query are using UDF, I think we can query "spark_plan" to get it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -91,6 +91,40 @@ object SnowflakeTelemetry {
...
 
+  /**
+    * Put the pushdown failure telemetry message to internal cache.
+    * The message will be sent later in batch.
+    *
+    * @param plan The logical plan to include the unsupported operations
+    * @param exception The pushdown unsupported exception
+    */
+  def addPushdownFailMessage(plan: LogicalPlan,
+                             exception: SnowflakePushdownUnsupportedException)
```
sfc-gh-bli, line 102: be aware of logical plan telemetry sometime fails to send because the String is too long.

sfc-gh-mrui: The telemetry messages are collected for statistics purpose. It doesn't matter if a small amount of messages are not sent successfully, I think.

PR #226, opened 2020-06-16
support feature branch scanning

sfc-gh-jfan:


whitesource/run_whitesource.sh
```diff
@@ -1,31 +1,39 @@
...
 cd ${WORKSPACE}
 
 # SCAN_DIRECTORIES is a comma-separated list (as a string) of file paths which contain all source code and build artifacts for this project
 SCAN_DIRECTORIES=$PWD
 
+# If your PROD_BRANCH is not master, you can define it here based on the need
+PROD_BRANCH="master"
+
 # PRODUCT_NAME is your team's name or overarching project name
 PRODUCT_NAME="spark-snowflake"
```
sfc-gh-mrui, line 21: I suspect GIT_BRANCH is not set. Could you please set it and echo it? 
export GIT_BRANCH=`git branch -l | grep “\*” | awk ‘NF>1{print $NF}’ | cut -d”)” -f1`
echo $GIT_BRANCH

PR #229, opened 2020-06-18
echo git branch name

sfc-gh-jfan:


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-169320

PR #233, opened 2020-06-24
SNOW-170484 Fix runQuery() to return a ResultSet W/O connection

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -412,18 +409,23 @@ object Utils {
...
+        .executeQuery(query)
+        .asInstanceOf[SnowflakeResultSet]
+        .getResultSetSerializables(Long.MaxValue)
     conn.close()
-    rs
+    if (resultSerializables.size() == 1) {
+      resultSerializables.get(0).getResultSet
+    } else {
+      throw new Exception("Fail to execute query by Utils.runQuery()")
+    }
```
sfc-gh-bli, line 422: Only support small result query？

sfc-gh-mrui: No. It supports all cases. "Long.MaxValue" is passed as expected partition size. It's sure only one object is returned.
Looks it is confusing, I will remove the THROW.

PR #235, opened 2020-06-26
SNOW-169389 Check missed files for COPY INTO TABLE.

sfc-gh-mrui:
If any missed files are detected, run additional COPY INTO TABLE command
to load missed files.


src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -680,6 +680,72 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
     }
```
sfc-gh-mrui, line 680: To reviewers:

Implementation description
====================
1. Introduce a new class FileUploadResult which includes fileName and fileSize. It is introduced to handle empty partitions if there are.

2. After spark executor finishes a partition, it returns a FileUploadResult object to spark driver

3. List of FileUploadResult are passed to the function to execute COPY command. Expected load files can generate from the list (empty files are skipped).

4. COPY with prefix is executed and check the result to find missed files.

5. If there are any missed files, execute the COPY with FILES clause for the missed files.

6. If there still are missed files, raise Exception. (Suppose it never happens)

Test description
============
1. Add two test flags: TH_COPY_INTO_TABLE_MISS_FILES_SUCCESSand TH_COPY_INTO_TABLE_MISS_FILES_FAIL

2. If TH_COPY_INTO_TABLE_MISS_FILES_FAIL is on, the first COPY command only loads parts of files, the 2nd COPY can load the missed files.

3. If both are on, after two COPY commands are executed, there still are missed files, Exception is raised.

For more details refer to the "Status Summary" for below Jira:
https://snowflakecomputing.atlassian.net/browse/SNOW-169389

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -380,8 +380,12 @@ object CloudStorageOperations {
     format: SupportedFormat = SupportedFormat.CSV,
     dir: Option[String] = None,
     compress: Boolean = true
-  )(implicit storage: CloudStorage): List[String] =
-    storage.upload(data, format, dir, compress)
+  )(implicit storage: CloudStorage): List[String] = {
+    val uploadResults = storage.upload(data, format, dir, compress)
+    val listBuffer = new ListBuffer[String]
+    uploadResults.foreach(x => listBuffer.add(x.fileName))
+    listBuffer.toList
```
sfc-gh-bli, line 387: You can replace these four lines by
```scala
storage.upload(data, format, dir, compress).map(_.fileName)
```

sfc-gh-mrui: Thanks! Follow it.

PR #237, opened 2020-07-02
SNOW-152613 Support microsecond when writing Timestamp to snowflake

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/Conversions.scala
```diff
@@ -89,6 +91,14 @@ private[snowflake] object Conversions {
     }
   }
```
sfc-gh-mrui, line 93: Spark only supports timestamp precision at most to micro second (don't support nano second). So it is impossible to support nano second. So SC is enhanced to support micro second.

SC has supported micro second when reading from SF with Arrow format. So reading from snowflake is not changed for this enhancement.

When writing data into snowflake, the Timestamp needs to be converted to String format and then SC runs COPY INTO TABLE to load data. Below two factors cause only milli seconds are used:
1. When converting Timestamp to String, SimpleDateFormat is used. Which is a legacy class and only supports milli second.
2. When running copy into table command, “TIMESTAMP_FORMAT='TZHTZM YYYY-MM-DD HH24:MI:SS.FF3'“ is specified. ”FF3” means milli second.

So the fix includes below parts:
1. Use a new class DateTimeFormatter when converting Timestamp to String. The target Timestamp format required timezone info. The spark local timezone is used.
2. Use “FF9“ to support more precision for COPY INTO TABLE: “TIMESTAMP_FORMAT='TZHTZM YYYY-MM-DD HH24:MI:SS.FF9'

PR #240, opened 2020-07-07
SNOW-173503 Don't send telemetry message: spark_plan

sfc-gh-mrui:


sfc-gh-mrui:
> If it is the finally decision, I am okey with it

It is decided on last team meeting.

PR #242, opened 2020-07-08
SNOW-173710 Leverage Spark retry-mechanism to handle throttling issue.

sfc-gh-mrui:
When the spark task fails, this task will be rescheduled by spark until the max attempt number is arrived. So we can leverage the spark retry instead of implementing retry in spark connector. Spark sleeps 3 seconds by default before reschedule. To handle cloud storage service throttling issue, it is necessary to use exponential sleep time (spark doesn’t support it). So, we introduce extra exponential sleep time based on the task’s attempt number.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
@@ -544,30 +544,9 @@ sealed trait CloudStorage {
   : SingleElementIterator = {
     val fileName = getFileName(partitionID, format, compress)
 
-    // When the upload task fails for non-fatal error, this task will be
-    // rescheduled by spark until the max attempt number is arrived.
-    // In some case, the task failure is caused by cloud service throttling.
-    // The default sleep time in spark (3 seconds) is not enough,
-    // so we do exponential backoff here.
-    val attemptNumber = TaskContext.get().attemptNumber()
-    var backoffInfo = "no_backoff"
-    if (attemptNumber > 0) {
-      // This part is tested manually.
-      // It is difficult to test with integration test because
-      // IT uses local cluster, for local cluster, maxTaskFailures
-      // is always 1. In debugger, manually set MAX_LOCAL_TASK_FAILURES
-      // in SparkContext can test the retry works.
-      val sleepTime = retrySleepTimeInMS(attemptNumber)
-      backoffInfo = s"""attemptNumber=$attemptNumber backoffTime=
-                       |${Utils.getTimeString(sleepTime)}
-                       |""".stripMargin.filter(_ >= ' ')
-      Thread.sleep(sleepTime)
-    }
-
     CloudStorageOperations.log.info(
       s"""${SnowflakeResultSetRDD.WORKER_LOG_PREFIX}:
          | Start writing partition ID:$partitionID as $fileName
-         | $backoffInfo
          |""".stripMargin.filter(_ >= ' '))
 
     // Either StorageInfo or fileTransferMetadata must be set.
@@ -658,28 +637,31 @@ sealed trait CloudStorage {
              |""".stripMargin.filter(_ >= ' ')
       }
 
-      // Unit Test code only
-      if (TestHook.isTestFlagEnabled(
-        TestHookFlag.TH_GCS_UPLOAD_RAISE_EXCEPTION)) {
-        if (attemptNumber > 1) {
-          TestHook.disableTestHook()
-        }
-        TestHook.raiseExceptionIfTestFlagEnabled(
-          TestHookFlag.TH_GCS_UPLOAD_RAISE_EXCEPTION,
-          "Negative test to raise error when uploading data to GCS"
-        )
-      }
+      // Unit Test code only. This part is tested manually.
+      // It is difficult to test with integration test because
+      // IT uses local cluster, for local cluster, maxTaskFailures
+      // is always 1. In debugger, manually set MAX_LOCAL_TASK_FAILURES
+      // in SparkContext can test the retry works.
+      TestHook.raiseExceptionIfTestFlagEnabled(
+        TestHookFlag.TH_GCS_UPLOAD_RAISE_EXCEPTION,
+        "Negative test to raise error when uploading data to GCS"
+      )
 
       CloudStorageOperations.log.info(
         s"""${SnowflakeResultSetRDD.WORKER_LOG_PREFIX}:
            | Finish writing partition ID:$partitionID $fileName
            | write row count is $rowCount.
            | Uncompressed data size is ${Utils.getSizeString(dataSize)}.
-           | $processTimeInfo $backoffInfo
+           | $processTimeInfo
            |""".stripMargin.filter(_ >= ' '))
     } catch {
       // Hit exception when uploading the file
       case e: Exception => {
+        val attemptNumber = TaskContext.get().attemptNumber()
+        val sleepTime = retrySleepTimeInMS(attemptNumber + 1)
+        // Sleep exponential time based on the attempt number.
+        Thread.sleep(sleepTime)
+
         val stringWriter = new StringWriter
         e.printStackTrace(new PrintWriter(stringWriter))
         val errmsg =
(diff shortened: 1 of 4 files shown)


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-173710

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -537,15 +544,30 @@ sealed trait CloudStorage {
...
-    // NOTE: For GCP, the data have to be cached and then uploaded.
-    val disable_cache_data = maxRetryCount <= 1 && storageInfo.isDefined
+    // When the upload task fails for non-fatal error, this task will be
+    // rescheduled by spark until the max attempt number is arrived.
+    // In some case, the task failure is caused by cloud service throttling.
+    // The default sleep time in spark (3 seconds) is not enough,
+    // so we do exponential backoff here.
+    val attemptNumber = TaskContext.get().attemptNumber()
+    var backoffInfo = "no_backoff"
+    if (attemptNumber > 0) {
```
sfc-gh-sshankar, line 554: Add the sleep at the end if the failure instead of the beginning of the second.

sfc-gh-mrui: OK.

sfc-gh-mrui: When I make the change, I found may it is better to check in the beginning. If we put the sleep in the end, when all retry fail, the last retry also needs to sleep. This sleep is useless.
It looks there is no way to get the MAX attempt number, if there is, we can avoid the last sleep.

sfc-gh-mrui: The possibility to hit the extra sleep is small. It's OK to put it in the end.

sfc-gh-sshankar: When is uploadRDD used vs upload ?
It might be cleaner to remove this sleep logic from uploadPartition -- it's kind of confusing to refer to taskcontext etc. in there. 
My suggestion: revert uploadPartition to roughly what it was as of this commit: 3bcdc53. That is let's leave the sleep/retry logic out of there.
Create a function wrapping uploadPartition that's passed in a task context and sleeps based on that. And lets make sure it's called from uploadRDD and upload

sfc-gh-mrui: uploadRDD is used for AWS and Azure. upload is used for GCP.

sfc-gh-mrui: uploadPartition() has implemented the retry logic from spark 2.7.0 for GCP support. From spark 2.7.1, we refactor uploadRDD() to use uploadPartition() to avoid duplicate code. After this refactor, AWS and Azure also uses uploadPartition.
I will do following as you suggested. It's sounds clearer.
1. Introduce a function called doUploadPartition() to read and upload data.
2. uploadPartition() call doUploadPartition() with try-catch and handle the sleep/telemetry message sending, etc

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -563,188 +559,158 @@ sealed trait CloudStorage {
...
-            TestHookFlag.TH_GCS_UPLOAD_RAISE_EXCEPTION,
-            "Negative test to raise error when uploading data to GCS"
-          )
+    try {
+      // Read data and upload to cloud storage
+      doUploadPartition(rows, format, compress, directory, partitionID,
+                        storageInfo, fileTransferMetadata)
+    } catch {
+      // Hit exception when uploading the file
+      case e: Exception => {
```
sfc-gh-sshankar, line 568: Do we want to do this on all exceptions or just throttling exceptions ?

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -563,188 +559,158 @@ sealed trait CloudStorage {
...
+        val attemptNumber = TaskContext.get().attemptNumber()
+        val sleepTime = retrySleepTimeInMS(attemptNumber + 1)
+        // Sleep exponential time based on the attempt number.
+        Thread.sleep(sleepTime)
+
+        val stringWriter = new StringWriter
+        e.printStackTrace(new PrintWriter(stringWriter))
+        val errmsg =
```
sfc-gh-sshankar, line 576: Let's log befgore we sleep and let's make sure the sleep time, attempt number and task ids are added  as part of the logs.

sfc-gh-mrui: Sure.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: We do this for all exceptions.

sfc-gh-mrui: Ideally, we should sleep only when throttling issue happens. But it is difficult to achieve because there are three kinds of cloud services and each one may raise multiple kinds of exceptions. And it doesn't deserve to spend much time here.

sfc-gh-sshankar: In the case of Nielsen, what specific exception was raised ? My concern is that this is too broad and screws up some fail-fast cases. 
If you do capture errors in such a broad manner, we should also have some param to turn off retries if needed.

sfc-gh-ema:
Am I correct in understanding that this retry mechanism is at the mercy of how Spark does task retry? I think this is a configurable parameter.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-ema: I agree, it seems that catching all exceptions is too general. We would incur extra, unnecessary sleep latency for other errors as well.

sfc-gh-mrui: I remember two kinds of exception is hit for Nielsen. One is something like "Sever is busy". I don't remember the other. Some month ago, another customer hit throttling issue on AWS, the error is only generic TIMEOUT.

By default, it retries 4 times. The average extra sleep time is (1.5 + 3 + 6 + 9) * 0.75 = 16 seconds.
Currently, for fail-fast cases, they configure spark spark.task.maxFailures as 1.It sleeps about 1 second.
It sounds the side-affect is not a big problem.

The retry count is controlled by spark, I think it may be not necessary to introduce new para on spark connector side.
But we can add a param to disable the extra sleep. Do you think it is necessary?

sfc-gh-mrui: I agree that it is pretty general to catch all exceptions. The another thing we need to consider is the task failure possibility. According to the OOB telemetry messages, only 2 upload errors are seen (in past 30 days). It can prove the task failure case is rare though the OOB message are sent only by SC 2.7.2 and 2.8.0.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -579,6 +582,7 @@ sealed trait CloudStorage {
              | partition ID:$partitionID $fileName
              | attemptNumber=$attemptNumber
              | backoffTime=${Utils.getTimeString(sleepTime)}
+             | useExponentialBackoff=$useExponentialBackoff
```
sfc-gh-sshankar, line 585: This log message with backOffTIme is confusing if we're not actually backing off.
Remove the backoffTime line from this message from here and add it to just above the Thread.sleep below.

PR #247, opened 2020-07-15
SNOW-170311 Send specific unsupported operation in generateQueries.

sfc-gh-mrui:
Before the enhancement, if generateQueries() meets unsupported
operation, it returns None, so a pushdown failure message for
"NoSuchElementException" is sent. It is unclear. With this fix,
specific info about the unsupported operation is sent.


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-170311

PR #248, opened 2020-07-16
SNOW-175528 investigated inset pushdown

sfc-gh-zli:


src/it/scala/net/snowflake/spark/snowflake/PushdownEnhancement01.scala
```diff
@@ -617,10 +600,60 @@ class PushdownEnhancement01 extends IntegrationSuiteBase {
...
+         |    "subquery_0"."name" in ('test4','test3','3','5','test2','test7','test1','6','test8','1','test6','4','2','test9')
+         |  )
+         |) as "subquery_1"
+         |""".stripMargin,
+      result,
+      expectedResult
+    )
+  }
+
+  test("test pushdown INSET() function with timestamps and double") {
```
sfc-gh-zli, line 618: To test the added default match in setToExpr

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
```diff
@@ -168,8 +170,21 @@ private[querygeneration] object MiscStatement {
...
       case d: Decimal => Literal(d, DecimalType(d.precision, d.scale))
-      case s: String => Literal(s, StringType)
+      case s @ (_:String | _:UTF8String) => Literal(s, StringType)
       case e: Expression => e
+      case default =>
+        try {
+          Literal(default)
+        }
+        catch {
+          case _: RuntimeException =>
```
sfc-gh-mrui, line 180: Some comments:
1. It sounds it is unclear what will happen for "Literal(default)".
    I prefer to specific more CASE for different types.
2. Are their specific reason to use "RuntimeException" instead of Exception?

sfc-gh-zli: 1. The code for Literal(default) is here: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/literals.scala#L56 
There are too many types, I don't want to copy them all into here. I thought about removing the previous cases and only use this default case, but that might introduce behavior change so I only added this new case. 
2. Yes using Exception to catch all error would be better.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
```diff
@@ -168,8 +171,16 @@ private[querygeneration] object MiscStatement {
   private final def setToExpr(set: Set[Any]): Seq[Expression] = {
     set.map {
       case d: Decimal => Literal(d, DecimalType(d.precision, d.scale))
-      case s: String => Literal(s, StringType)
+      case s @ (_:String | _:UTF8String) => Literal(s, StringType)
+      case d: Double => Literal(d, DoubleType)
       case e: Expression => e
+      case default =>
```
sfc-gh-mrui, line 177: 1. Use format like: s"${default.getClass.toString} @ MiscStatement. setToExpr"
2. Add comment before Throw like:
            // This exception is not a real issue. It will be caught in
            // QueryBuilder.treeRoot and a telemetry message will be sent if
            // there are any snowflake tables in the query.

PR #249, opened 2020-07-17
SNOW-175796 Send unsupported function name for aggregate function

sfc-gh-mrui:
Currently, telemetry message with "aggregateexpression' is sent
for unsupported aggregate function. It's unclear.


sfc-gh-mrui:
The integration test on GCP hang because timeout happens when downloading result chunk from GCS. It looks a test env issue. The test on Travis and my laptop works well. Let's ignore it so far.

PR #250, opened 2020-07-18
SNOW-96700 Add pushdown test for BasicStatement.

sfc-gh-mrui:


sfc-gh-mrui:
The integration test on GCP hang because timeout happens when downloading result chunk from GCS. It looks a test env issue. The test on Travis and my laptop works well. Let's ignore it so far.

PR #257, opened 2020-07-24
SNOW-96700 Add pushdown test for NumericStatement.

sfc-gh-mrui:


src/it/scala/net/snowflake/spark/snowflake/PushdownEnhancement02.scala
```diff
@@ -455,6 +457,58 @@ class PushdownEnhancement02 extends IntegrationSuiteBase {
...
+      testPushdown(
+        s"""SELECT ( ROUND ( "SUBQUERY_0"."D1" , 0 ) ) AS "SUBQUERY_1_COL_0",
+           |( ROUND ( "SUBQUERY_0"."F1" , 0 ) ) AS "SUBQUERY_1_COL_1",
+           |( 3.141592653589793 ) AS "SUBQUERY_1_COL_2" FROM
+           |( SELECT * FROM ( $test_table_number ) AS
+           |"SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0"
+           |""".stripMargin,
+        resultDF,
+        expectedResult
+      )
```
sfc-gh-zli, line 503: Maybe we can change this testPushdown function to take regex matches as ground truth, so that if there is uncertainties in the executed query like random(), we can use regex to match it. Just my random thoughts, maybe there is not much use cases for the effort.

sfc-gh-mrui: It's good idea. But so far, it is no necessary. Let's do it, when necessary.

PR #258, opened 2020-07-27
SNOW-178870 Don't pushdown SORT to snowflake.

sfc-gh-mrui:


sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-178870

sfc-gh-mrui:
We defer this change to future because:
1. The possibility to hit this issue is extremely low.
2. Even if the issue happens, it can be workaround easily by disabling pushdown or set partition_size_by_mb to larger value. 3. So far, we didn't hear any customer report it.
4. It may cause some customers to concern perf regression if they depends on SORT pushdown.

PR #260, opened 2020-07-29
SNOW-179168 Simplify the pushdown for SORT + LIMIT

sfc-gh-mrui:


sfc-gh-mrui:
> LGTM, generally. Do we know what the impact of using `SortLimitQuery` with the `orderExpr` is, compared to the way that it is now? Does this affect what we can push down/performance?

It doesn't affect what can be pushdown. Its performance may be better or equal based on how SF optimizes the query.

The actual code change is to removed below two cases:
CASE_1:   case Limit(limitExpr, Sort(orderExpr, true, _))
CASE_2:  case Sort(orderExpr, true, Limit(limitExpr, _))

For CASE_1: it is after "case Limit(limitExpr, _)", so it is never used.
For CASE_2: there is "case Sort(orderExpr, true, _)" after it. The only difference is that there is no redundant "LIMIT X".

PR #263, opened 2020-07-30
SNOW-175818 Support pushdown regexp_replace()

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/StringStatement.scala
```diff
@@ -92,6 +92,24 @@ private[querygeneration] object StringStatement {
           null
         }
 
+      case RegExpReplace(subject, regexp, rep) =>
+        def unEscape(stmt: SnowflakeSQLStatement): String = {
```
sfc-gh-bli, line 96: `unEscape` -> `unescape`

PR #264, opened 2020-08-03
SNOW-177689 Enable IT for GCP

sfc-gh-mrui:
The IT test on GCP is disabled because environment issue. Now, it looks the env issue has been resolved. re-enable it.


sfc-gh-ema (approved):
What was the environment issue?

sfc-gh-mrui:
> What was the environment issue?
The environment issue looks like GitHub blocks the traffic to GCP and now they unblock the traffic.

The symptom we can saw is that the chunk file download timeout when the test reads large result set.
This happens from one day suddenly (it is sure there is no PR merged on that day).
I retried periodically. And I found IT test on GCP works from yesterday. So, I think we can re-enable it.

sfc-gh-bli (approved):
LGTM

sfc-gh-sshankar (approved):
Approval

PR #267, opened 2020-08-06
SNOW-180368 Add format check for unix_timestamp() pushdown

sfc-gh-mrui:
Perform a format check for pushdown unix_timestamp() on string column.
Only enable the pushdown if the format is supported by Snowflake.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/QueryBuilder.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/QueryBuilder.scala
@@ -23,6 +23,7 @@ import org.apache.spark.sql.catalyst.plans._
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.execution.datasources.LogicalRelation
 import org.apache.spark.sql.types.{StructField, StructType}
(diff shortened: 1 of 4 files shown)


src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/QueryBuilder.scala
```diff
@@ -39,6 +40,11 @@ private[querygeneration] class QueryBuilder(plan: LogicalPlan) {
     */
   private var foundSnowflakeRelation = false
 
+  /**
+    * Logger for pushdown processing
+    */
+  private[querygeneration] val log = LoggerFactory.getLogger(getClass)
```
sfc-gh-bli, line 46: It is ok for now.
I think we should use Spark Logging instead of Logger. it is a big project, many codes should be refactored.

sfc-gh-mrui: I think the SC and spark already shared the same logging file.

sfc-gh-bli: Yes, they do. I mean we can simplify the interface

PR #279, opened 2020-08-20
SNOW-185140 Add unit test for Misc items

sfc-gh-mrui:
1. Remove some unused legacy functions
2. Add some unit test for some uncovered code


src/it/scala/net/snowflake/spark/snowflake/IntegrationSuiteBase.scala
```diff
@@ -116,6 +116,15 @@ trait IntegrationSuiteBase
     resultOptions
   }
```
sfc-gh-mrui, line 118: This function is moved from SnowflakeRelationSuite.scala.

src/main/scala/net/snowflake/spark/snowflake/CloudCredentialsUtils.scala
```diff
@@ -26,35 +26,9 @@ import net.snowflake.client.jdbc.internal.amazonaws.auth.{
   InstanceProfileCredentialsProvider
 }
 import net.snowflake.client.jdbc.internal.microsoft.azure.storage.StorageCredentialsSharedAccessSignature
-import net.snowflake.spark.snowflake.Parameters.MergedParameters
 import org.apache.hadoop.conf.Configuration
-import org.apache.spark.sql.SQLContext
 
 private[snowflake] object CloudCredentialsUtils {
```
sfc-gh-mrui, line 31: Some explain for reviewers:
Three functions are removed because:
1. They are not used.
2. They looks to be used to specify info for external stage for COPY command.
     Even with existing external stage, these info is not needed. 
     Actually, We create a named stage for the external stage,
     and we use the named stage directly in the COPY command.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
```diff
@@ -57,56 +57,6 @@ private[snowflake] object FilterPushdown {
     keepNameCase: Boolean
   ): Option[SnowflakeSQLStatement] = {
```
sfc-gh-mrui, line 59: Explain to reviewers:
These two functions is moved out of buildFilterStatement() so that I can add some unit test for them.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -670,31 +670,6 @@ object Parameters {
       */
     def postActions: Array[String] = parameters(PARAM_POSTACTIONS).split(";")
```
sfc-gh-mrui, line 672: Some explain to reviewers:
These two functions are not used. And they are in Parameter. I don't think any customer should use them.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -160,75 +160,6 @@ object Utils {
   def makeTempPath(tempRoot: String): String =
     Utils.joinUrls(tempRoot, UUID.randomUUID().toString)
```
sfc-gh-mrui, line 162: Explain to reviewers:
These two functions are not used. Firstly, I want to keep them since they are in Utils. But these two functions look like legacy functions for external stage. I think we'd better to remove them.

PR #282, opened 2020-08-27
SNOW-186380 Support table name with special character by quotation

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -331,7 +331,7 @@ private[io] object StageWriter {
     val table = params.table.get
     val tempTable =
       TableName(
-        s"${table.name.replace('"', '_')}_staging_${Math.abs(Random.nextInt()).toString}"
+        s"${table.name.replaceAll("[^a-zA-Z0-9_]", "X")}_staging_${Math.abs(Random.nextInt()).toString}"
```
sfc-gh-bli, line 334:
```suggestion
        s"${table.name.replaceAll("\\W", "X")}_staging_${Math.abs(Random.nextInt()).toString}"
```

sfc-gh-bli: [^a-zA-Z0-9_] == \w
https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html

sfc-gh-mrui: Thanks! Change it.

sfc-gh-bli: You can click on `Commit Suggestion` button to commit the suggestion.

PR #284, opened 2020-08-31
SNOW-188466 test high degree of concurrency

sfc-gh-zli:
Added a new GitHub Action test to test high degree of concurrency. In total there are 5 worker node, each with 2 cores and 1300MB memory. We submit 10 jobs in parallel each requires 1 core and 600MB memory. Each job will upload 200 partitions. Those jobs should finish at the same time. Given that GitHub Action's machine only have 2 cores and 7G memory, this is the simplest solution.


sfc-gh-mrui:
@sfc-gh-zli Do we still need this PR?

sfc-gh-zli:
Close this PR because there is no need to add such test any more.

PR #287, opened 2020-09-09
SNOW-188512 added test to simulate throttling

sfc-gh-zli:


.github/docker/build_image.sh
```diff
@@ -33,7 +33,7 @@ cd ../..
 
 # Build docker image
 docker build \
---build-arg SPARK_URL=https://apache.osuosl.org/spark/spark-3.0.0/spark-3.0.0-bin-hadoop2.7.tgz \
+--build-arg SPARK_URL=https://archive.apache.org/dist/spark/spark-3.0.0/spark-3.0.0-bin-hadoop2.7.tgz \
```
sfc-gh-zli, line 36: previous link has expired

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -723,6 +723,12 @@ sealed trait CloudStorage {
       "Negative test to raise error when uploading data to GCS"
     )
 
+    TestHook.raiseExceptionIfTestFlagEnabledWithMaxCount(
+      TestHookFlag.TH_UPLOAD_RAISE_EXCEPTION_WITH_COUNT,
+      "Negative test to raise error when uploading data to GCS",
+      2
+    )
```
sfc-gh-zli, line 730: Throw exception for the first two times, then pass.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -723,6 +723,14 @@ sealed trait CloudStorage {
       "Negative test to raise error when uploading data to GCS"
     )
 
+    // When attempt number is smaller than 2, throw exception
```
sfc-gh-mrui, line 726: It looks it is no necessary to keep two flags for this case. Can TH_GCS_UPLOAD_RAISE_EXCEPTION and TH_UPLOAD_RAISE_EXCEPTION_WITH_COUNT merged as one?

sfc-gh-zli: Merged two flags into one

sfc-gh-mrui (approved):
Thanks for the good work!

PR #289, opened 2020-09-15
Updated test readme

sfc-gh-zli:


README-TEST.md
```diff
@@ -169,3 +169,36 @@ Files in ClusterTest
 * src/main/scala//net/snowflake/spark/snowflake/ClusterTest.scala: The driver class to run test
 * src/main/scala//net/snowflake/spark/snowflake/testsuite/ClusterTestSuiteBase.scala: Test case base class.
```
sfc-gh-mrui, line 170: Please add the two new test cases file name here and short description.

README-TEST.md
```diff
@@ -169,3 +169,36 @@ Files in ClusterTest
...
 * Files in src/main/scala: the Scala test related files.
+
+## Specific test case description
+### Low memory test
+Test uploading large partition with a small amount of memory is defined in Cluster Test. The partition used in the test is of size 227 MB after compression and 330 MB before compression. By limiting the size of executor memory, multiple OOM errors can be reproduced.
+
+The first one is Garbage Collect OOM, which happens when the executor does not have enough memory to execute Java GC. Then the connector hit OOM while trying to write data into a buffer. The processed size column indicates the processed data size before OOM. The connector hits close OOM when it tries to copy the buffer into a byte array.
```
sfc-gh-mrui, line 177: I think "Garbage Collect OOM" is confusing the reader. It should happen when generating test data. I suggest you update the test case not to generate random data for each row. For example, we generate some random values in the begin, and use them later. We don't care the compression ration. It is good enough to make the compressed size to be equivalent.

sfc-gh-mrui (changes requested):

README-TEST.md
sfc-gh-zli: Nice catch, thanks!

README-TEST.md
sfc-gh-zli: Not generating random data does save runtime memory. But I think it's hard to measure how much runtime heap memory it actually saved. To me this Garbage Collect OOM basically means Spark does not have enough reserved memory to execute normal functionality. I think the purpose of this test is already achieved, so I tend to leave the test as is for now.

PR #297, opened 2020-10-02
SNOW-201485: Port over Snowpark-style integration with Spark Datafram…

sfc-gh-ema:
This changeset ports over the Snowpark-style integration with some of the Dataframe suites in Spark SQL. 

We can un-blacklist some of the tests that work with the connector (more of them should work with this than with Snowpark). This work is tracked in [SNOW-201512](https://snowflakecomputing.atlassian.net/browse/SNOW-201512).

Also, there may be one bug discovered by running these tests. This is tracked in [SNOW-201486](https://snowflakecomputing.atlassian.net/browse/SNOW-201486).


src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
```diff
@@ -0,0 +1,288 @@
...
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable
+import scala.io.Source
+
+/**
+  * Base class for writing integration tests which run against a real Snowflake cluster.
+  */
+trait IntegrationEnv
```
sfc-gh-mrui, line 38: 1. I guess your intention is to reorg "IntegrationSuiteBase" to extract some common code for existing IT test and the new ID test. It is good idea. Could you confirm you did extra change in the same time?

Below is the original traits for IntegrationSuiteBase. I see "FunSuite" is new. Is it intended?

trait IntegrationSuiteBase
    extends QueryTest
    with Matchers
    with BeforeAndAfterAll	
    with BeforeAndAfterEach {

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -643,4 +645,10 @@ object Utils {
         println(s"Fail to print result set: ${e.getMessage}")
     }
   }
+
+  /**
+   * Generate random suffixes; useful for use
+   * in generating collision-safe identifiers in tests.
+   */
+  def randomSuffix: String = Math.abs(Random.nextLong()).toString
```
sfc-gh-mrui, line 717: I think randomSuffix is for testing purpose. if so, can you add it to test TestUtils.

src/it/scala/org/apache/spark/sql/SFDataFrameAggregateSuite.scala
```diff
@@ -0,0 +1,89 @@
+package org.apache.spark.sql
```
sfc-gh-mrui, line 1: I know your general idea is to port SnowPark test code to Spark connector and you should add some change to make it works for spark connector. I think it will help the review if you can point out what's your change.

build.sbt
```diff
@@ -61,6 +61,9 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
       // "org.apache.spark" %% "spark-sql-kafka-0-10" % "2.4.0",
       "org.apache.spark" %% "spark-core" % testSparkVersion % "provided, test",
       "org.apache.spark" %% "spark-sql" % testSparkVersion % "provided, test",
+      "org.apache.spark" %% "spark-core" % testSparkVersion % "provided, test" classifier "tests",
```
sfc-gh-mrui, line 64: I think the snow park is based on spark 3.0. But spark connector needs to support spark 2.3 and 2.4.
Do we need to port these test for spark 2.3/2.4? I think it may not be necessary. If we have to, it sounds a lot work for the porting.

src/it/scala/org/apache/spark/sql/snowflake/SFTestWrapperSparkSession.scala
```diff
@@ -0,0 +1,37 @@
...
+    extends SparkSession(sc) {
+
+  override def createDataFrame[A <: Product: TypeTag](data: Seq[A]): DataFrame = {
+    // Create a DataFrame from the Seq using Spark's normal way.
+    val origDF = super.createDataFrame(data)
+
+    // Write this DataFrame to Snowflake, in effect testing the connector's write-out
+    // code path, and then read it back as a Snowflake DataFrame. This uses
+    // the temporary schema defined in the run properties, tempTestSchema
+    val tableName = s"tempTestTable_${Utils.randomSuffix}"
```
sfc-gh-sshankar, line 19: This is nice, but your tests could take a long time to run. Consider caching the table name based on data, you are likely to see a lot of repeated use of the same sequence.

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+class SFDataFrameWindowFunctionsSuite
+    extends DataFrameWindowFunctionsSuite
+    with SFTestSessionBase
+    with SFQueryTest
+    with SFTestData {
+
+  override def spark: SparkSession = getSnowflakeSession()
+
+  override protected def blackList: Seq[String] =
+    Seq(
```
sfc-gh-sshankar, line 14: Why would you blacklist *anything* ? Unlike TS, the idea here is for us to push what's possible and execute the rest in Spark, correct ?

sfc-gh-sshankar (commented):
I don't understand why anything is blacklisted.

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
sfc-gh-ema: Yeah, as I mentioned, I'd like to remove from the blacklist tests that work here:
https://snowflakecomputing.atlassian.net/browse/SNOW-201512

I still see some failures when I try to run the whole suite, and there's a large overlap between what fails and what is on this SnowPark-inherited blacklist, so they'll need to be investigated.

For now, I see this blacklist as still having a use: to make sure there aren't any failures in our integration tests while we patch things up.

src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
sfc-gh-ema: Yeah, it's needed for `BeforeAndAfterAll` and `BeforeAndAfterEach`; it was otherwise provided originally by `QueryTest`.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
sfc-gh-ema: Moved.

build.sbt
sfc-gh-ema: I'll test the changes with the other versions.

src/it/scala/org/apache/spark/sql/snowflake/SFTestWrapperSparkSession.scala
sfc-gh-ema: Hmm, compared to copying the data over to Snowflake, I don't think generating a random string is going to use up all that much time. Admittedly, the write() and read() add a bunch of time to the testing, and this happens when the Dataframe is accessed for the first time in a suite. I don't have any other bright ideas for optimizing this.

sfc-gh-mrui (approved):
LGTM. My understanding is that this PR is to introduce the framework. We will enable more test cases later.

build.sbt
sfc-gh-bli: aha, you figure out how to import test jar on SBT.

sfc-gh-bli: Yes, that is a problem. We should decide to run such test only against spark 3.0 or all supported version (2.3 and 2.4). We may be not able to share test since they do change a lot.

src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
```diff
@@ -0,0 +1,288 @@
+/*
+ * Copyright 2015-2016 Snowflake Computing
```
sfc-gh-bli, line 2: 2015 - 2016? not 2020?

src/it/scala/org/apache/spark/sql/SFDataFrameAggregateSuite.scala
```diff
@@ -0,0 +1,89 @@
...
+    extends DataFrameAggregateSuite
+    with SFTestSessionBase
+    with SFQueryTest
+    with SFTestData {
+
+  override def spark: SparkSession = getSnowflakeSession()
+
+  override protected def blackList: Seq[String] =
+    Seq(
+      // contains RDD, replaced by TS - groupBy
```
sfc-gh-bli, line 15: I believe you should modify these black lists

src/it/scala/org/apache/spark/sql/SFDataFrameRangeSuite.scala
```diff
@@ -0,0 +1,17 @@
+package org.apache.spark.sql
+
+import org.apache.spark.sql.snowflake.{SFQueryTest, SFTestSessionBase}
+
+class SFDataFrameRangeSuite extends DataFrameRangeSuite with SFTestSessionBase with SFQueryTest {
+  // replace Spark Utils
```
sfc-gh-bli, line 6: Does our connector support Range function?

sfc-gh-bli:
My major concern is, we should make sure all tests run on snowflake but not on spark.

src/it/scala/org/apache/spark/sql/SFDataFrameAggregateSuite.scala
sfc-gh-ema: This will be addressed in a later PR.

sfc-gh-ema:
To be clear: this PR addresses only porting over the test framework from Snowpark. There are some follow-up tasks from this PR: 

1. Un-blacklisting tests from the suites (and possibly removing the blacklist altogether)
2. Fixing the aggregate pushdown issue
3. Adding more test suites

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -39,6 +39,8 @@ import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.spark.sql.types.{StructField, StructType}
 import org.slf4j.LoggerFactory
 
+import scala.util.Random
```
sfc-gh-sshankar, line 42: Revert

src/test/scala/net/snowflake/spark/snowflake/TestUtils.scala
```diff
@@ -23,6 +23,8 @@ import java.util.Calendar
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types._
 
+import scala.util.Random
```
sfc-gh-sshankar, line 26: revert ?

sfc-gh-bli (approved):

PR #298, opened 2020-10-08
Add limit 1 for aggregates with no grouping expressions

sfc-gh-ema:
This applies the 'limit 1' fix for aggregate expressions pushed down in queries where there is no grouping expression.

Removing the blacklisted item from SFDataFrameAggregateSuite effectively adds a test for this item.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/SnowflakeQuery.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/SnowflakeQuery.scala
@@ -212,7 +212,9 @@ case class AggregateQuery(columns: Seq[NamedExpression],
       ConstantString("GROUP BY") +
         mkStatement(groups.map(expressionToStatement), ",")
     } else {
-      EmptySnowflakeSQLStatement()
+      // SNOW-201486: Insert a limit 1 to ensure that only one row returns if there
+      // are no grouping expressions
+      ConstantString("LIMIT 1").toStatement
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/SnowflakeQuery.scala
```diff
@@ -212,7 +212,9 @@ case class AggregateQuery(columns: Seq[NamedExpression],
       ConstantString("GROUP BY") +
         mkStatement(groups.map(expressionToStatement), ",")
     } else {
-      EmptySnowflakeSQLStatement()
+      // SNOW-201486: Insert a limit 1 to ensure that only one row returns if there
+      // are no grouping expressions
+      ConstantString("LIMIT 1").toStatement
```
sfc-gh-sshankar, line 217: Do we ever generate having classes ? If so, this may lead to incorrect syntax.

sfc-gh-ema: We do not generate `HAVING` clauses. These are always converted into `WHERE` clauses during planning.

sfc-gh-sshankar (approved):
Approval

sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-233617

sfc-gh-mrui:
https://snowflakecomputing.atlassian.net/browse/SNOW-201486

PR #300, opened 2020-10-09
SNOW-202025: Add DateFunctionsSuite to IT Tests

sfc-gh-ema:
This PR adds `DateFunctionsSuite` to the connector's IT tests. The suite is more-or-less a direct copy of Spark's DateFunctionsSuite, with the `Seq.toDF()` implicit conversion overridden to create a Snowflake Dataframe.

There are 8 failures in the test discovered by porting these tests over. These are tracked in: 
https://snowflakecomputing.atlassian.net/browse/SNOW-205533

For now, these tests are disabled in `SFDateFunctionsSuite`.


src/it/scala/org/apache/spark/sql/SFDateFunctionsSuite.scala
```diff
@@ -0,0 +1,925 @@
...
+import org.apache.spark.SparkUpgradeException
+import org.apache.spark.sql.catalyst.util.DateTimeTestUtils.{CEST, LA}
+import org.apache.spark.sql.catalyst.util.DateTimeUtils
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.snowflake.{SFQueryTest, SFTestSessionBase}
+import org.apache.spark.sql.types.DoubleType
+import org.apache.spark.unsafe.types.CalendarInterval
+
+class SFDateFunctionsSuite extends SFQueryTest with SFTestSessionBase {
```
sfc-gh-mrui, line 35: Is this test suite new developed or ported from SNOW-PARK? If it is new, I think they should be put in net.snowflake...

sfc-gh-ema: It's inherited. Also, it extends directly from some of the Spark test classes.

sfc-gh-mrui (approved):
LGTM. All changes are to enable more tests.

src/it/scala/org/apache/spark/sql/snowflake/SnowflakeSqlImplicits.scala
```diff
@@ -0,0 +1,243 @@
...
+import scala.collection.Map
+import scala.language.implicitConversions
+import scala.reflect.runtime.universe.TypeTag
+
+/**
+ * A collection of implicit methods for converting common Scala objects into [[org.apache.spark.sql.Dataset]]s.
+ *
+ * @since 1.6.0
+ */
+abstract class SnowflakeSQLImplicits(sfSession: SFTestWrapperSparkSession)
```
sfc-gh-sshankar, line 31: Why do we need to copy these over ? If we are extending session, don't the implicits just apply ?

sfc-gh-ema: Because of the way implicits work in Scala, if I inherited the superclass test suite, the implicits defined *there* would apply, instead of anything I tried to override. So instead, I need to copy/paste all the test code from the suite that this copies.

I also couldn't just subclass `SqlImplicits`, because the method would override the `Seq -> DatasetHolder` implicit conversion defined there, and `DatasetHolder` is a case class, which couldn't be subclassed: meaning that the `Seq` had to be converted into `DatasetHolder`, and `DatasetHolder` is effectively final.

That's how I settled on copy/pasting the implicits definitions instead of using any inheritance.

sfc-gh-sshankar (approved):

PR #301, opened 2020-10-12
SNOW-206210 Decrease log volume for ported snowpark test

sfc-gh-mrui:


sfc-gh-ema:
Do we really want to only change the logging level only for the ported test suites? That is, are we actually getting any value for the other log messages in our IT environment?

src/it/scala/org/apache/spark/sql/snowflake/SFTestSessionBase.scala
```diff
@@ -30,6 +31,8 @@ trait SFTestSessionBase extends IntegrationEnv {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
+    // Run ported test in WARN level to avoid large log volume
+    reconfigureLogFile(PORTED_TEST_LOG4J_PROPERTY)
```
sfc-gh-ema, line 35: Why not just use `Logger.setLevel()`?

src/it/resources/ported_test_log4j.properties
```diff
@@ -0,0 +1,42 @@
+#
```
sfc-gh-ema, line 1: Do we actually need to do this complex reconfiguration, rather than just downgrade the log levels?

src/it/scala/org/apache/spark/sql/snowflake/SFTestSessionBase.scala
sfc-gh-mrui: Logger.setLevel() is NOT a static function. It needs to be called for a logger object. So it can't change logging level for all loggers.

src/it/resources/ported_test_log4j.properties
sfc-gh-mrui: By default, IT uses the spark default log property file in org/apache/spark/log4j-defaults.properties in spark core jar file e.g. spark-core_2.12-3.0.0.jar. We couldn't change it.
In practice, I save a copy at "src/it/resources/log4j_default.properties". This config file only changes 'log4j.rootCategory'  from "INFO, console" to "WARN, console".
When we need to change the log level, we use this logging property,
After the test, we configure back to the default.

sfc-gh-mrui:
> Do we really want to only change the logging level only for the ported test suites? That is, are we actually getting any value for the other log messages in our IT environment?

Before the ported test suites are merged, the log size is about 2.5M, the travis log limit is 4M. So, I think we can keep the logging for existing suites for debugging purpose.

src/it/scala/org/apache/spark/sql/snowflake/SFTestSessionBase.scala
sfc-gh-ema: `Configurator.setAllLevels()` 

Ref: https://stackoverflow.com/questions/23434252/programmatically-change-log-level-in-log4j2

sfc-gh-mrui: That page is for log4j2 which is different to log4j.
I have tried more research. I didn't find better way to set log level in program.

sfc-gh-ema (approved):

PR #302, opened 2020-10-15
SNOW-206588: Fix translations of null literals

sfc-gh-ema:
Previously, NULLs were translated incorrectly in value-typed literals (e.g., Ints, Floats, etc.), as they would be cast, and therefore default to some value when they should instead be null.

This PR adds some safeties for null literals, and re-enables the two tests that were previously failing.


src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
```diff
@@ -63,24 +63,24 @@ private[snowflake] object FilterPushdown {
...
               .replace("'", "''")
               .replace("\\", "\\\\")
-          ) !
+            )) !
         case DateType =>
-          StringVariable(value.asInstanceOf[Date].toString) + "::DATE"
+          StringVariable(Option(value).map(_.asInstanceOf[Date].toString)) + "::DATE"
         case TimestampType =>
-          StringVariable(value.asInstanceOf[Timestamp].toString) + "::TIMESTAMP(3)"
+          StringVariable(Option(value).map(_.asInstanceOf[Timestamp].toString)) + "::TIMESTAMP(3)"
```
sfc-gh-mrui, line 73: It looks it will cause some precision loss when timestamp is converted to "::TIMESTAMP(3)". I know it is not related to this fix. But I think it needs to be investigated.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
```diff
@@ -63,24 +63,24 @@ private[snowflake] object FilterPushdown {
       dataType match {
```
sfc-gh-mrui, line 63: buildValueWithType() and buildValue() are internal functions. I believe many of the cases are not test-covered. I suggest you move out these 2 functions and make them as "private[snowflake]" and add unit test to cover them.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
```diff
@@ -91,19 +91,19 @@ private[snowflake] object FilterPushdown {
...
-        case x: Int => IntVariable(x) !
-        case x: Long => LongVariable(x) !
-        case x: Short => ShortVariable(x) !
-        case x: Boolean => BooleanVariable(x) !
-        case x: Float => FloatVariable(x) !
-        case x: Double => DoubleVariable(x) !
-        case x: Byte => ByteVariable(x) !
-        case _ => ConstantString(value.toString) !
+        case x: Date => StringVariable(Option(x).map(_.toString)) + "::DATE"
+        case x: Timestamp => StringVariable(Option(x).map(_.toString)) + "::TIMESTAMP(3)"
```
sfc-gh-mrui, line 98: As as above "::TIMESTAMP(3)" may cause precision loss.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -19,31 +19,24 @@
...
 
 import net.snowflake.client.jdbc.telemetry.{Telemetry, TelemetryClient}
-
-import scala.concurrent.{Await, ExecutionContext, Future}
-import scala.concurrent.duration.Duration
-import scala.util.Try
-import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry
-import org.apache.spark.sql.types._
-import org.slf4j.LoggerFactory
+import net.snowflake.spark.snowflake.DefaultJDBCWrapper.DataBaseOperations
```
sfc-gh-mrui, line 28: In general, I suggest not to change the "import" exception necessary.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -715,9 +708,9 @@ private[snowflake] class SnowflakeSQLStatement(
     bindVariableEnabled: Boolean
   )(implicit conn: Connection): ResultSet =
     if (bindVariableEnabled) executeWithBindVariable(conn)
-    else executeWithoutBindVaribale(conn)
+    else executeWithoutBindVariable(conn)
```
sfc-gh-mrui, line 711: I don't see any difference for line 711/713. Why they are diff?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -767,23 +760,47 @@ private[snowflake] class SnowflakeSQLStatement(
...
         element match {
           case ele: StringVariable =>
-            statement.setString(index + 1, ele.variable)
+            if (ele.variable.isDefined)
+              statement.setString(index + 1, ele.variable.get)
+            else
+              statement.setNull(index + 1, java.sql.Types.VARCHAR)
           case ele: Identifier =>
-            statement.setString(index + 1, ele.variable)
+              statement.setString(index + 1, ele.variable.get)
```
sfc-gh-mrui, line 768: There is no "if (ele.variable.isDefined)" check for Identifier. Is it expected?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -767,23 +760,47 @@ private[snowflake] class SnowflakeSQLStatement(
       case (element, index) =>
         element match {
           case ele: StringVariable =>
-            statement.setString(index + 1, ele.variable)
+            if (ele.variable.isDefined)
```
sfc-gh-mrui, line 763: Since NULL is handled in executeWithBindVariable(), I think there may be no test coverage for them.
I suggest to add some unit test for it. Actually, we can add unit test in IT, so we can have a real connection to make the unit test easier. I have some test for stage in StageSuite.scala. FYI.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -852,6 +869,16 @@ private[snowflake] object EmptySnowflakeSQLStatement {
   def apply(): SnowflakeSQLStatement = new SnowflakeSQLStatement()
 }
 
+private[snowflake] object ConstantStringVal {
+  def apply(l: Any): StatementElement = {
+    if (l == null || l.toString == "null") {
```
sfc-gh-mrui, line 874: Do we need to consider " null", "Null", etc?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
```diff
@@ -78,29 +78,28 @@ private[querygeneration] object BasicStatement {
...
+              StringVariable(Some(l.toString())) ! // else "'" + str + "'"
             }
           case DateType =>
             ConstantString("DATEADD(day,") + IntVariable(
-              l.value.asInstanceOf[Int]
+              Option(l.value).map(_.asInstanceOf[Int])
             ) +
               ", TO_DATE('1970-01-01'))" // s"DATEADD(day, ${l.value}, TO_DATE('1970-01-01'))"
           case TimestampType =>
             ConstantString("to_timestamp_ntz(") + l.toString() + ", 6)"
```
sfc-gh-mrui, line 92: Open question: NULL is considered for before and after CASE, but TimestampType is not changed. Is it intended?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: UnitTestWithConnectionSuite should be a better reference.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-ema: I fixed a typo. "Varibale" vs "Variable".

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-ema: Yes, it's expected. Identifier() does not take an option, because it's expected that it will never be None for Identifiers.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-ema: Good point. I am simply porting over old behavior, but we should do that. Will make the change.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
sfc-gh-ema: I don't understand this question. Could you clarify?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: Good catch.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
sfc-gh-mrui: What I mean is for before and after CASES, you use Option(l.value).map(_.asInstanceOf[TYPE_NAME]). But, for TimestampType, we don't use this format.
Now, I understand, the before and after, Option(l.value) is needed if asInstanceOf[] is used.
This is resoved.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-ema: The `FilterPushdown` object is already marked as `private[snowflake]`. These are 'static' functions.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-ema: Yeah, this does seem suspect. Good catch. Let's investigate this, but outside of this PR.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-ema: scalafmt formatter changed these. :(

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-mrui: What I mean is these two functions are defined in function buildFilterStatement() not in class FilterPushdown. To test them with unit test, they should be moved out as FilterPushdown's member functions.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-mrui: agree.

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-ema: So, I discovered that these functions require the variable `schema`, which is passed down through the top-level function `buildFilterStatement`, so I can't move these out of `buildFilterStatement` without refactoring. Since this is invoked from `SnowflakeRelation`, I would prefer to not perform extensive refactoring at this time.

sfc-gh-mrui (approved):
LGTM.

sfc-gh-sshankar (approved):
Approval

PR #303, opened 2020-10-16
SNOW-206836 Send telemetry message for failed query.

sfc-gh-mrui:
Refer to Jira description for details.
https://snowflakecomputing.atlassian.net/browse/SNOW-206836

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
@@ -607,6 +607,70 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
     }
   }
 
+  // Negative test for read fail.
+  // The query tries to convert a String to number, but the value is not number
+  // So the reading data frame hits runtime error.
+  test("testFailToRead") {
+    setupStringBinaryTable
+
+    assertThrows[Exception]({
+      sparkSession.read
+        .format(SNOWFLAKE_SOURCE_NAME)
+        .options(connectorOptionsNoTable)
+        .option("query", s"select to_number(v) from $test_table_string_binary")
+        .load()
+        .collect()
+    })
+  }
+
+  test("test Write table with AUTOINCREMENT column") {
+    // Create a table with AUTOINCREMENT column
+    jdbcUpdate(
+      s"""create or replace table $test_table_write (
+         | int_c int, c_string string(1024), id_c int AUTOINCREMENT (100, 10))
+         | """.stripMargin)
+
+    // Write 5 rows without data for AUTOINCREMENT column
+    val df1 = sparkSession.read
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("query", "select seq4() as INT_C, 'test123' as C_STRING from table(generator(rowcount => 5))")
+      .load()
+    df1.write
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("dbtable", test_table_write)
+      .option("columnmap", "Map(INT_C -> INT_C, C_STRING -> C_STRING)")
+      .mode(SaveMode.Append)
+      .save()
+
+    // Write 5 rows with data for AUTOINCREMENT column
+    val df2 = sparkSession.read
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("query", "select seq4(), 'test456', 1234 from table(generator(rowcount => 5))")
+      .load()
+    df2.write
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("dbtable", test_table_write)
+      .mode(SaveMode.Append)
+      .save()
+
+    // Verify result for AUTOINCREMENT column
+    val expectedAnswer = Array(
+      Row(100), Row(110), Row(120), Row(130), Row(140),
+      Row(1234), Row(1234), Row(1234), Row(1234), Row(1234)
+    )
+    val result = sparkSession.read
+      .format(SNOWFLAKE_SOURCE_NAME)
+      .options(thisConnectorOptionsNoTable)
+      .option("query", s"select ID_C from $test_table_write order by ID_C")
+      .load()
+
+    checkAnswer(result, expectedAnswer)
+  }
+
   test("test read write StringBinary") {
     setupStringBinaryTable
     // COPY UNLOAD can't be run because it doesn't support binary
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -28,8 +28,8 @@ import net.snowflake.client.jdbc.telemetry.{Telemetry, TelemetryClient}
 import net.snowflake.spark.snowflake.DefaultJDBCWrapper.DataBaseOperations
 import net.snowflake.spark.snowflake.Parameters.MergedParameters
 import net.snowflake.spark.snowflake.Utils.JDBC_DRIVER
+import net.snowflake.client.jdbc.{SnowflakeResultSet, SnowflakeStatement}
 import net.snowflake.spark.snowflake.io.SnowflakeResultSetRDD
-import org.apache.spark.sql.catalyst.expressions.Literal
 import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry
 import org.apache.spark.sql.types._
 import org.slf4j.LoggerFactory
@@ -691,6 +691,10 @@ private[snowflake] class SnowflakeSQLStatement(
 
   private val log = LoggerFactory.getLogger(getClass)
 
+  private var lastQueryID: String = _
+
+  private[snowflake] def getLastQueryID(): String = lastQueryID
+
   def +(element: StatementElement): SnowflakeSQLStatement =
     new SnowflakeSQLStatement(numOfVar + element.isVariable, element :: list)
 
@@ -729,7 +733,16 @@ private[snowflake] class SnowflakeSQLStatement(
     log.info(s"$logPrefix $query")
 
     val statement = conn.prepareStatement(query)
-    DefaultJDBCWrapper.executePreparedQueryInterruptibly(statement)
+    try {
+      val rs = DefaultJDBCWrapper.executePreparedQueryInterruptibly(statement)
+      lastQueryID = rs.asInstanceOf[SnowflakeResultSet].getQueryID
+      rs
+    } catch {
+      case th: Throwable => {
+        lastQueryID = statement.asInstanceOf[SnowflakeStatement].getQueryID
+        throw th
+      }
+    }
   }
 
   private def executeWithBindVariable(conn: Connection): ResultSet = {
@@ -808,8 +821,16 @@ private[snowflake] class SnowflakeSQLStatement(
         }
     }
 
-    DefaultJDBCWrapper.executePreparedQueryInterruptibly(statement)
-
+    try {
+      val rs = DefaultJDBCWrapper.executePreparedQueryInterruptibly(statement)
+      lastQueryID = rs.asInstanceOf[SnowflakeResultSet].getQueryID
+      rs
+    } catch {
+      case th: Throwable => {
+        lastQueryID = statement.asInstanceOf[SnowflakeStatement].getQueryID
+        throw th
+      }
+    }
   }
 
   override def equals(obj: scala.Any): Boolean =
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
@@ -82,7 +82,7 @@ object SnowflakeTelemetry {
                        success: Boolean,
                        useProxy: Boolean,
                        queryID: Option[String],
-                       errorThrow: Option[Throwable]): Unit =
+                       throwable: Option[Throwable]): Unit =
   {
     val metric: ObjectNode = mapper.createObjectNode()
     metric.put(TelemetryOOBFields.SPARK_CONNECTOR_VERSION, Utils.VERSION)
@@ -94,11 +94,11 @@ object SnowflakeTelemetry {
     metric.put(TelemetryOOBFields.SUCCESS, success)
     metric.put(TelemetryOOBFields.USE_PROXY, useProxy)
     metric.put(TelemetryOOBFields.QUERY_ID, queryID.getOrElse("NA"))
-    if (errorThrow.isDefined) {
-      metric.put(TelemetryOOBFields.EXCEPTION_CLASS_NAME, errorThrow.get.getClass.toString)
-      metric.put(TelemetryOOBFields.EXCEPTION_MESSAGE, errorThrow.get.getMessage)
+    if (throwable.isDefined) {
+      metric.put(TelemetryOOBFields.EXCEPTION_CLASS_NAME, throwable.get.getClass.toString)
+      metric.put(TelemetryOOBFields.EXCEPTION_MESSAGE, throwable.get.getMessage)
       val stringWriter = new StringWriter
-      errorThrow.get.printStackTrace(new PrintWriter(stringWriter))
+      throwable.get.printStackTrace(new PrintWriter(stringWriter))
       metric.put(TelemetryOOBFields.EXCEPTION_STACKTRACE, stringWriter.toString)
     } else {
       metric.put(TelemetryOOBFields.EXCEPTION_CLASS_NAME, "NA")
@@ -194,7 +194,7 @@ object SnowflakeTelemetry {
                                          queryId: String,
                                          queryStatus: String,
                                          elapse: Long,
-                                         errorThrow: Option[Throwable],
+                                         throwable: Option[Throwable],
                                          details: String
                                          ): Unit = {
     val metric: ObjectNode = mapper.createObjectNode()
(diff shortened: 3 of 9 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -188,6 +188,31 @@ object SnowflakeTelemetry {
...
+                                         elapse: Long,
+                                         message: String,
+                                         details: String
+                                         ): Unit = {
+    val metric: ObjectNode = mapper.createObjectNode()
+    metric.put(TelemetryQueryResultFields.SPARK_CONNECTOR_VERSION, Utils.VERSION)
+    metric.put(TelemetryQueryResultFields.OPERATION, operation)
+    metric.put(TelemetryQueryResultFields.QUERY, query)
+    metric.put(TelemetryQueryResultFields.QUERY_ID, queryId)
+    metric.put(TelemetryQueryResultFields.QUERY_RESULT, queryResult)
```
sfc-gh-ema, line 205: queryResult and message are independent fields, but it seems that message is only used for exception conditions. Do you ever expect 'success', but also with a populated message field?

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -468,66 +480,118 @@ private[io] object StageWriter {
...
+        // Only load part of missed files.
+        // Exception is raised for the failure.
+        val secondCopyFileSet: Option[mutable.Set[String]] =
+        if (TestHook.isTestFlagEnabled(
+          TestHookFlag.TH_COPY_INTO_TABLE_MISS_FILES_FAIL)) {
+          Some(missedFileSet.grouped(2).toList.head)
+        } else {
+          Some(missedFileSet)
+        }
+        progressMessage.append(
```
sfc-gh-ema, line 522: What if we hit an issue that's not caught, i.e., there is a hang, or the process ends unexpectedly? Then we don't get any of these progress messages. Should we think about explicitly sending the messages as the progress is updated?

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -607,6 +607,70 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
     }
   }
 
+  // Negative test for read fail.
```
sfc-gh-ema, line 610: Do these tests actually test any of the telemetry functionality?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -711,13 +712,38 @@ private[snowflake] class SnowflakeSQLStatement(
 
   def isEmpty: Boolean = list.isEmpty
 
-  def execute(
-    bindVariableEnabled: Boolean
-  )(implicit conn: Connection): ResultSet =
-    if (bindVariableEnabled) executeWithBindVariable(conn)
-    else executeWithoutBindVaribale(conn)
+  private[snowflake] def executeWithoutThrow(
```
sfc-gh-ema, line 715: Why not just send telemetry in `catch` instead of testing for `isDefined` for the third tuple field?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: Yes. I intend to make it to be easy to support success query.  In the beginning, the message is named as "spark_query_fail". I change it to "spark_query_result" with the QUERY_RESULT field.
The high level idea is for SC dashboard. If we send success message for read and write. we can monitor each customer to see how many spark job they run, how many fail/succeed, and look into details for the fail reason.
We can discuss in the later meeting.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: Good catch. We catch "Throwable", so the possibility not to print the message is rare. But it is possible, for example, the task is running and the user kills it.
I will use a string array for the progress. The messages are logged directly. If telemetry message need to be send, all details information is included.

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -705,12 +705,38 @@ private[snowflake] class SnowflakeSQLStatement(
...
+  : (ResultSet, String, Option[Throwable]) =
+    if (bindVariableEnabled) {
+      executeWithBindVariable(conn)
+    } else {
+      executeWithoutBindVariable(conn)
+    }
 
-  private def executeWithoutBindVariable(conn: Connection): ResultSet = {
+  private def executeWithoutBindVariable(conn: Connection)
+  : (ResultSet, String, Option[Throwable]) = {
```
sfc-gh-sshankar, line 739: Create a case class for this return type.

sfc-gh-sshankar: Also, this is not a good way to write this.
```
case class ResultInfo(queryID: String, result: Option[ResultSet])
private def executeWithoutBindVariable(conn): ResultInfo
```
And use Try/Success/Failure to deal with exceptions.
```
def execute() = {
  val executeResult = Try(...)
}
```

sfc-gh-sshankar: Or better yet, just use try/catch.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -188,6 +188,31 @@ object SnowflakeTelemetry {
     )
   }
 
+  private[snowflake] def sendQueryResult(conn: Connection,
```
sfc-gh-sshankar, line 191: is this sending the query result or the query status ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -198,9 +223,36 @@ object TelemetryTypes extends Enumeration {
...
+  // Spark connector version
+  val SPARK_CONNECTOR_VERSION: String = "spark_connector_version"
+  // The unsupported operation for pushdown
+  val OPERATION: String = "operation"
+  // Query statement
+  val QUERY: String = "query"
+  // query ID if available
+  val QUERY_ID: String = "queryid"
+  // Query result: fail vs success
+  val QUERY_RESULT: String = "result"
```
sfc-gh-sshankar, line 247: result != status

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -198,9 +223,36 @@ object TelemetryTypes extends Enumeration {
...
+  // query execution time
+  val QUERY_ELAPSE: String = "elapse"
```
sfc-gh-sshankar, line 249: elapsed_time

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -198,9 +223,36 @@ object TelemetryTypes extends Enumeration {
...
+  // The error message for the exception
+  val EXCEPTION_MESSAGE: String = "message"
```
sfc-gh-sshankar, line 251: exception_message

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
```diff
@@ -36,13 +36,31 @@ private[snowflake] object StageReader {
...
       statement,
       s"@$stage/$prefix/",
       compressFormat,
       format
-    ).execute(params.bindVariableEnabled)(conn)
+    )
+
+    val startTime = System.currentTimeMillis()
+    val (res, queryID, errorThrow) =
+      copyStatement.executeWithoutThrow(params.bindVariableEnabled)(conn)
```
sfc-gh-sshankar, line 49: Why do we need this, aren't we logging telemetry in execute ?

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -468,66 +482,115 @@ private[io] object StageWriter {
...
-      )
-
-      def getMissedFileInfo(missedFileSet: mutable.Set[String]): String = {
-        s"""missedFileCount=${missedFileSet.size}
-           | Files: (${missedFileSet.mkString(", ")})
-           |""".stripMargin.filter(_ >= ' ')
+    var lastQueryID: String = null
+    try {
+      // execute the COPY INTO TABLE statement
+      val (resultSet, queryID, errorThrow) =
```
sfc-gh-sshankar, line 491: This is super-ugly, you shouldn't have to change this stuff,

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -468,66 +482,115 @@ private[io] object StageWriter {
...
+      // Get missed files if there are any.
+      var missedFileSet = getCopyMissedFiles(params, resultSet, expectedFileSet)
 
-      // It is expected all the files must be loaded.
+      // If any files are missed, execute 2nd COPY command
       if (missedFileSet.nonEmpty) {
-        throw new SnowflakeConnectorException(
-          s"""These files are missed when COPY INTO TABLE:
-             | ${getMissedFileInfo(missedFileSet)}
+        // Negative test:
```
sfc-gh-sshankar, line 513: This seems unrelated.

sfc-gh-sshankar (commented):
I think this needs more work.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: The reason to work in this way to retrieve the queryID for query fail.
For successful query, we can get the queryID from resultSet. For failed query, we have to get the queryID from the Statement. But the statement is created in side of executeWithoutBindVariable()/executeWithBindVariable().
To get the queryID and not to introduce duplicate code, I change the internal function to return the Exception and the caller processes the exception.

But I agree it is good idea to use CASE CLASS for the return type. I plan to use ExecuteResult.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: query status. I will change the function name as sendQueryStatus()

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: will update

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: Will change:
TelemetryQueryResultFields  -> TelemetryQueryStatusFields
QUERY_RESULT -> QUERY_STATUS

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: will update

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
sfc-gh-mrui: executeWithoutThrow() doesn't log error inside.
This is to sending a telemetry message when reading with COPY INTO LOCATION fails.
The telemetry message is independent to send log entries to telemetry.

PS: I am a little confusing about the question. If this doesn't address your comment, could you please explain the question?

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: It's some complex, but I think it is very useful for diagnosis.
I can just change to call executeWithoutThrow(), check error and send telemetry message.
That will be easier, but I think the current is better.
With current change, if COPY fails, we can get details info about the progress and situation.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: This is not new code. I guess it makes confused because the format change causes it to be like new code.

src/main/scala/net/snowflake/spark/snowflake/io/StageReader.scala
sfc-gh-mrui: Oh, I see. I think you mean we can move to send the telemetry message inside of executeWithoutThrow().
It's doable, I didn't do that because some information for the telemetry message are collected by the caller.
If we want to send message inside of executeWithoutThrow(), we may add some extra parameters.
In the future, we will send successful messages too, some info are not ready in the executeWithoutThrow() function such as total data size, partition count, etc.
So, I think it should be better to process it by caller and keep executeWithoutThrow() simple.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: After considering it again, I change it back to "message".
The reason is that we have already sent Exception with OOB telemetry message on executors. I think we'd better  use the same field name for exception. Otherwise, it may confusing when querying the message from snowhouse.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -691,6 +691,10 @@ private[snowflake] class SnowflakeSQLStatement(
 
   private val log = LoggerFactory.getLogger(getClass)
 
+  private var lastQueryID: String = _
```
sfc-gh-sshankar, line 694: This should be a thread local, no ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -82,7 +82,7 @@ object SnowflakeTelemetry {
                        success: Boolean,
                        useProxy: Boolean,
                        queryID: Option[String],
-                       exception: Option[Exception]): Unit =
+                       errorThrow: Option[Throwable]): Unit =
```
sfc-gh-sshankar, line 85: errorThrow => throwable

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -198,9 +230,40 @@ object TelemetryTypes extends Enumeration {
...
+
+object TelemetryQueryStatusFields {
+  // Spark connector version
+  val SPARK_CONNECTOR_VERSION: String = "spark_connector_version"
+  // The query operation: read vs write
+  val OPERATION: String = "operation"
+  // Query statement
+  val QUERY: String = "query"
+  // query ID if available
+  val QUERY_ID: String = "queryid"
```
sfc-gh-sshankar, line 252: query_id. Please have consistent style

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-sshankar: This is only set on failure. I think it would be useful to always set it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: will update

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: yes. It has set up for both failure and success. For example,
      val rs = DefaultJDBCWrapper.executePreparedQueryInterruptibly(statement)
      lastQueryID = rs.asInstanceOf[SnowflakeResultSet].getQueryID
      rs

sfc-gh-mrui: SnowflakeSQLStatement is defined as "class SnowflakeSQLStatement", not "object SnowflakeSQLStatement". The caller doesn't share SnowflakeSQLStatement instance between thread. So it is not necessary to make it thread local.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: I will use QUERYID to align OOB message.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -249,7 +249,7 @@ object TelemetryQueryStatusFields {
   // Query statement
   val QUERY: String = "query"
   // query ID if available
-  val QUERY_ID: String = "queryid"
+  val QUERYID: String = "queryid"
```
sfc-gh-sshankar, line 252: query_id, to match max_retry and use_proxy, etc.

sfc-gh-mrui: Will update

sfc-gh-sshankar (approved):

PR #304, opened 2020-10-21
SNOW-208934 Log spark configuration

sfc-gh-mrui:
In general, below configuration information are collected.
1. JVM memory and cores configuration
2. Spark options
3. Spark task info

The configuration are logged in spark log.
The configuration for #1 #2 are also included in the existing telemetry message "spark_client_info" for diagnosis purpose.

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
@@ -73,12 +73,29 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
   }
 
   test("inject exceptions for Arrow read") {
+    // Configure some spark options
+    // manually check these options are sent correctly.
+    val separateSpark = SparkSession.builder
+      .master("local")
+      .appName("test config info sent")
+      .config("spark.sql.shuffle.partitions", "6")
+      .config("spark.driver.cores", "3")
+      .config("spark.driver.memory", "2G")
+      .config("spark.driver.memoryOverhead", "456M")
+      .config("spark.executor.memory", "888M")
+      .config("spark.executor.pyspark.memory", "123M")
+      .config("spark.executor.memoryOverhead", "444M")
+      .config("spark.driver.extraJavaOptions", s"-Duser.timezone=GMT")
+      .config("spark.executor.extraJavaOptions", s"-Duser.timezone=UTC")
+      .config("spark.sql.session.timeZone", "America/Los_Angeles")
+      .getOrCreate()
+
     try {
       if (!params.useCopyUnload) {
         // Enable test hook to simulate error when closing a result set on driver.
         // This exception doesn't affect the final result
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_DRIVER_FAIL_CLOSE_RESULT_SET)
-        sparkSession.read
+        separateSpark.read
           .format(SNOWFLAKE_SOURCE_NAME)
           .options(connectorOptionsNoTable)
           .option("dbtable", s"$test_table_basic")
@@ -88,7 +105,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when opening a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_OPEN_RESULT_SET)
         assertThrows[Exception]({
-          sparkSession.read
+          separateSpark.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
@@ -99,7 +116,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when reading a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_READ_RESULT_SET)
         assertThrows[Exception]({
-          sparkSession.read
+          separateSpark.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
--- a/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
@@ -75,7 +75,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
   test("inject exceptions for Arrow read") {
     // Configure some spark options
     // manually check these options are sent correctly.
-    val separateSpark = SparkSession.builder
+    val sparkSessionWithMoreOptions = SparkSession.builder
       .master("local")
       .appName("test config info sent")
       .config("spark.sql.shuffle.partitions", "6")
@@ -95,7 +95,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when closing a result set on driver.
         // This exception doesn't affect the final result
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_DRIVER_FAIL_CLOSE_RESULT_SET)
-        separateSpark.read
+        sparkSessionWithMoreOptions.read
           .format(SNOWFLAKE_SOURCE_NAME)
           .options(connectorOptionsNoTable)
           .option("dbtable", s"$test_table_basic")
@@ -105,7 +105,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when opening a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_OPEN_RESULT_SET)
         assertThrows[Exception]({
-          separateSpark.read
+          sparkSessionWithMoreOptions.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
@@ -116,7 +116,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when reading a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_READ_RESULT_SET)
         assertThrows[Exception]({
-          separateSpark.read
+          sparkSessionWithMoreOptions.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
--- a/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
@@ -73,29 +73,12 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
   }
 
   test("inject exceptions for Arrow read") {
-    // Configure some spark options
-    // manually check these options are sent correctly.
-    val sparkSessionWithMoreOptions = SparkSession.builder
-      .master("local")
-      .appName("test config info sent")
-      .config("spark.sql.shuffle.partitions", "6")
-      .config("spark.driver.cores", "3")
-      .config("spark.driver.memory", "2G")
-      .config("spark.driver.memoryOverhead", "456M")
-      .config("spark.executor.memory", "888M")
-      .config("spark.executor.pyspark.memory", "123M")
-      .config("spark.executor.memoryOverhead", "444M")
-      .config("spark.driver.extraJavaOptions", s"-Duser.timezone=GMT")
-      .config("spark.executor.extraJavaOptions", s"-Duser.timezone=UTC")
-      .config("spark.sql.session.timeZone", "America/Los_Angeles")
-      .getOrCreate()
-
     try {
       if (!params.useCopyUnload) {
         // Enable test hook to simulate error when closing a result set on driver.
         // This exception doesn't affect the final result
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_DRIVER_FAIL_CLOSE_RESULT_SET)
-        sparkSessionWithMoreOptions.read
+        sparkSession.read
           .format(SNOWFLAKE_SOURCE_NAME)
           .options(connectorOptionsNoTable)
           .option("dbtable", s"$test_table_basic")
@@ -105,7 +88,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when opening a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_OPEN_RESULT_SET)
         assertThrows[Exception]({
-          sparkSessionWithMoreOptions.read
+          sparkSession.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
@@ -116,7 +99,7 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
         // Enable test hook to simulate error when reading a result set.
         TestHook.enableTestFlagOnly(TestHookFlag.TH_ARROW_FAIL_READ_RESULT_SET)
         assertThrows[Exception]({
-          sparkSessionWithMoreOptions.read
+          sparkSession.read
             .format(SNOWFLAKE_SOURCE_NAME)
             .options(connectorOptionsNoTable)
             .option("dbtable", s"$test_table_basic")
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -199,9 +199,9 @@ private[snowflake] class JDBCWrapper {
            |""".stripMargin.filter(_ >= ' '))
     }
 
-    val snowflakeClientInfo = Utils.getClientInfoString()
-    log.info(snowflakeClientInfo)
-    System.setProperty("snowflake.client.info", snowflakeClientInfo)
+    // Important: Set "snowflake.client.info" is very important!
+    // For more details, refer to PROPERTY_NAME_OF_CONNECTOR_VERSION
+    System.setProperty("snowflake.client.info", Utils.getClientInfoString())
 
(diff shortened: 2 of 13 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -152,6 +152,9 @@ private[snowflake] case class SnowflakeRelation(
   // without first executing it.
   private def getRDD[T: ClassTag](statement: SnowflakeSQLStatement,
                                   resultSchema: StructType): RDD[T] = {
+    log.info(s"${SnowflakeResultSetRDD.MASTER_LOG_PREFIX} System config with " +
+      s"read: ${SnowflakeTelemetry.getSystemConfigWithoutTaskInfo().toPrettyString}")
+
```
sfc-gh-mrui, line 157: To reviewers:
Log system config for all READ spark job.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -88,7 +91,8 @@ object SnowflakeTelemetry {
                        success: Boolean,
                        useProxy: Boolean,
                        queryID: Option[String],
-                       throwable: Option[Throwable]): Unit =
+                       throwable: Option[Throwable],
+                       config: Option[ObjectNode]): Unit =
```
sfc-gh-mrui, line 95: To Reviewers:
When exception occurs for executors, OOB message is sent. The system config is included in the telemetry message.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -201,7 +208,8 @@ object SnowflakeTelemetry {
                                          queryStatus: String,
                                          elapse: Long,
                                          throwable: Option[Throwable],
-                                         details: String
+                                         details: String,
+                                         config: Option[ObjectNode]
```
sfc-gh-mrui, line 212: To Reviewers:
When exception occurs for driver when executing query, telemetry message is sent. The system config is included in the telemetry message.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -219,13 +227,154 @@ object SnowflakeTelemetry {
...
+      System.getProperty(TelemetryConstValues.JVM_PROPERTY_NAME_OS_NAME))
+    val rt = Runtime.getRuntime
+    metric.put(TelemetryConfigFields.MAX_MEMORY_IN_MB, rt.maxMemory() / MB)
+    metric.put(TelemetryConfigFields.TOTAL_MEMORY_IN_MB, rt.totalMemory() / MB)
+    metric.put(TelemetryConfigFields.FREE_MEMORY_IN_MB, rt.freeMemory() / MB)
+    metric.put(TelemetryConfigFields.CPU_CORES, rt.availableProcessors())
+    metric
+  }
+
+  // Below specifies the spark options may affects spark connector
```
sfc-gh-mrui, line 293: To reviewers:
Actually, we could collect all spark options. But, I spend time to go through spark options to identify the options which may affect SC. The reason is that I find some spark options are credential. It may be a problem if they are logged or sent to telemetry.
The disadvantage is that some options may be missed.

Please advice any others spark option needs to be collected. Or any options are not good to collect.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -71,8 +71,11 @@ object Utils {
   /**
     * Client info related variables
     */
-  private[snowflake] lazy val sparkAppName =
-    SparkContext.getOrCreate().getConf.get("spark.app.name", "")
+  private[snowflake] lazy val sparkAppName = if (SparkEnv.get != null) {
```
sfc-gh-mrui, line 74: To reviewers:
SparkContext is not available on executor. So, it is changed to use SparkEnv.

src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
```diff
@@ -73,12 +73,29 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
   }
 
   test("inject exceptions for Arrow read") {
+    // Configure some spark options
+    // manually check these options are sent correctly.
+    val separateSpark = SparkSession.builder
```
sfc-gh-sshankar, line 78: If it's getOrCreate, how do you know a new one will be created ?

sfc-gh-sshankar: Also I don't understand the name separateSpark

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -199,9 +199,9 @@ private[snowflake] class JDBCWrapper {
            |""".stripMargin.filter(_ >= ' '))
     }
 
-    val snowflakeClientInfo = Utils.getClientInfoString()
-    log.info(snowflakeClientInfo)
-    System.setProperty("snowflake.client.info", snowflakeClientInfo)
+    // Important: Set "snowflake.client.info" is very important!
```
sfc-gh-sshankar, line 202: Why is it important ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-sshankar: Why not send config via in-band telemetry and just once ? This seems like a lot of repetition. I guess it's just for convenience of querying ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -111,6 +115,9 @@ object SnowflakeTelemetry {
       metric.put(TelemetryOOBFields.EXCEPTION_MESSAGE, "NA")
       metric.put(TelemetryOOBFields.EXCEPTION_STACKTRACE, "NA")
     }
+    if (config.isDefined) {
+      metric.set(TelemetryOOBFields.SYSTEM_CONFIG, config.get)
```
sfc-gh-sshankar, line 119: Use config.foreach

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-sshankar: Also, why pass this in for each call. The config is static, correct ?

src/main/scala/net/snowflake/spark/snowflake/SparkConnectorContext.scala
```diff
@@ -0,0 +1,43 @@
...
+
+object SparkConnectorContext {
+  // TODO: need to merge with another PR to use LoggerWrapper.
+  import org.slf4j.LoggerFactory
+  private[snowflake] val logger = LoggerFactory.getLogger(getClass)
+
+  private var isExecutorConfigLogged = false
+
+  // One executor may execute multiple tasks, it is enough to log once.
+  private[snowflake] def executorLogConfigIfNotYet(): Unit = {
```
sfc-gh-sshankar, line 32: You should rename this function.
recordConfig.
Also, why not just use this same call to store the telemetry version of the config you want to send.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: This setting will make the Snowflake GS to know the connection is from Spark Connector. In some special case, GS will do some special handling.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: With “config.foreach”, the spark option will be in current level (not a sub level). When we query the message, we may see spark option names are delimited by “.”, SC option names are delimited “_”. Does this concern you?

sfc-gh-mrui: I had misunderstanding for config.foreach. I will update it.

src/main/scala/net/snowflake/spark/snowflake/SparkConnectorContext.scala
sfc-gh-mrui: I will rename it.
Could you please explain below comments?
"Also, why not just use this same call to store the telemetry version of the config you want to send."

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: I tried to use “config.foreach”, but it doesn’t work in this case. I think it is nice to have. Let’s just use “IF”.
config.foreach(println(_)) // This works
config.foreach(x => metric.set(TelemetryQueryStatusFields.SYSTEM_CONFIG, x)) // This raises exception.

The exception is “net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode cannot be cast to scala.runtime.Nothing$”

src/main/scala/net/snowflake/spark/snowflake/SparkConnectorContext.scala
sfc-gh-mrui: This log is for executor only. I rename it as recordConfigOnExecutor() finally.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-ema: Does it make more sense to have a blacklist instead of a whitelist here?

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -71,8 +71,11 @@ object Utils {
...
+    SparkEnv.get.conf.get("spark.app.name", "")
```
sfc-gh-ema, line 75: Is this enough to uniquely identify a Spark run? Do we expect to use this combined with timestamps?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: The difficulty is how to build a complete blacklist.
Before building the whitelist, I considered to skip spark options whose name includes "key"/"password"/"credential" (ignore case).
But I think whitelist is safer, so I give up that proposal.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
sfc-gh-mrui: This is only a general information field. It is not unique.
We'll use "spark.app.id" as the unique id lately.

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/SparkConnectorContext.scala
```diff
@@ -0,0 +1,42 @@
...
+import net.snowflake.spark.snowflake.io.SnowflakeResultSetRDD.WORKER_LOG_PREFIX
+import org.slf4j.LoggerFactory
+
+object SparkConnectorContext {
+  private[snowflake] val logger = new LoggerWithTelemetry(LoggerFactory.getLogger(getClass))
+
+  private var isExecutorConfigLogged = false
+
+  // One executor may execute multiple tasks, it is enough to log once.
+  private[snowflake] def recordConfigOnExecutor(): Unit = {
```
sfc-gh-sshankar, line 31: Just synchronize() this function

sfc-gh-mrui: I will update. I also found it is a problem to synchronize on to-be-updated value. I will introduce an independent locker "Object" for synchronization.

sfc-gh-ema (approved):

src/it/scala/net/snowflake/spark/snowflake/InjectTestSuite01.scala
```diff
@@ -73,12 +73,29 @@ class InjectTestSuite01 extends IntegrationSuiteBase {
   }
 
   test("inject exceptions for Arrow read") {
+    // Configure some spark options
+    // manually check these options are sent correctly.
```
sfc-gh-sshankar, line 77: Manually check ?

sfc-gh-mrui: @sfc-gh-sshankar The message is sent to telemetry in snowhouse. Currently, there is no way to login to snowhouse in IT test. So, we can only do manual test.

sfc-gh-mrui: @sfc-gh-sshankar After consideration, I think we can use Unit test instead of integration test to test spark options are included in the telemetry message. I have updated the PR. Could you please take a look?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -228,6 +230,143 @@ object SnowflakeTelemetry {
     )
     SnowflakeTelemetry.send(conn.getTelemetry)
   }
+
+  // Configuration retrieving is optional for for diagnostic purpose,
+  // so it never raises exception.
+  private[snowflake] def getSystemConfigWithoutTaskInfo(): ObjectNode = {
```
sfc-gh-sshankar, line 236: Why do we have functions with and without task info ? Can't you just always include the task info if it's present ?

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -670,8 +673,10 @@ object Utils {
   }
 
   def getClientInfoJson(): ObjectNode = {
-    val metric: ObjectNode = mapper.createObjectNode()
+    SnowflakeTelemetry.getSystemConfigWithoutTaskInfo()
```
sfc-gh-sshankar, line 676: Is this only called on the driver now ?

src/main/scala/net/snowflake/spark/snowflake/SparkConnectorContext.scala
sfc-gh-sshankar: Why not just call this recordConfig ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -228,6 +230,143 @@ object SnowflakeTelemetry {
...
+    addSparkConfig(metric)
+  }
+
+  // Configuration retrieving is optional for for diagnostic purpose,
+  // so it never raises exception.
+  private[snowflake] def getSystemConfigWithTaskInfo(): ObjectNode = {
+    val metric = getSystemConfigWithoutTaskInfo()
+
+    try {
+      addSystemConfigInternal(metric)
```
sfc-gh-sshankar, line 263: Inline the call here. You don;t need this to be a separate function.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -228,6 +230,143 @@ object SnowflakeTelemetry {
...
+    Utils.addVersionInfo(metric)
+    // Add JVM system config
+    addJVMConfig(metric)
+    // Add Spark option
+    addSparkConfig(metric)
+  }
+
+  // Configuration retrieving is optional for for diagnostic purpose,
+  // so it never raises exception.
+  private[snowflake] def getSystemConfigWithTaskInfo(): ObjectNode = {
```
sfc-gh-sshankar, line 259: getClientConfig()
{
 addversioninfo
addjvmconfig
addsparkconfig
addtaskinfo
}

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -228,6 +230,143 @@ object SnowflakeTelemetry {
...
+    val rt = Runtime.getRuntime
+    metric.put(TelemetryClientInfoFields.MAX_MEMORY_IN_MB, rt.maxMemory() / MB)
+    metric.put(TelemetryClientInfoFields.TOTAL_MEMORY_IN_MB, rt.totalMemory() / MB)
+    metric.put(TelemetryClientInfoFields.FREE_MEMORY_IN_MB, rt.freeMemory() / MB)
+    metric.put(TelemetryClientInfoFields.CPU_CORES, rt.availableProcessors())
+    metric
+  }
+
+  // Below specifies the spark options may affects spark connector
+  val AFFECTED_SPARK_OPTIONS = Set(
```
sfc-gh-sshankar, line 286: rename to sparkOptions

sfc-gh-sshankar: (affected is misleading, the connector doesn't change these options)

sfc-gh-sshankar (commented):
Could be simplified

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -51,6 +53,7 @@ object SnowflakeTelemetry {
   def sendClientInfoTelemetryIfNotYet(extraValues: Map[String, String],
                                       conn: Connection): Unit = {
     if (!hasClientInfoSent) {
+      SparkConnectorContext.recordConfig()
```
sfc-gh-sshankar, line 56: What is the difference between recordConfig and sendClientInfoTelemetry

sfc-gh-sshankar (approved):
Approved, but I have a question about recordConfig and add telemetry if needed. We should clean that up in a separate PR.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: recordConfig() is supposed to be called for both executor and driver to log configuration.
sendClientInfoTelemetry() is supposed to be called once on spark driver to send a spark_client_info telemetry message.

For spark driver, recordConfig() can be called in other places, but I think it is a good place to call it in sendClientInfoTelemetry().

PR #305, opened 2020-10-23
SNOW-211001 Make telemetry message to use consistent property name

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -100,10 +106,6 @@ object SnowflakeTelemetry {
       val stringWriter = new StringWriter
       throwable.get.printStackTrace(new PrintWriter(stringWriter))
       metric.put(TelemetryOOBFields.EXCEPTION_STACKTRACE, stringWriter.toString)
-    } else {
```
sfc-gh-mrui, line 103: To reviewers:
Removing this ELSE clause doesn't cause any actual change. Currently "throwable.isDefined" is true for all callers.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -234,7 +236,43 @@ object TelemetryTypes extends Enumeration {
...
+  val APPLICATION_NAME = "application_name"
+  val SCALA_VERSION = "scala_version"
+  val JAVA_VERSION = "java_version"
+  val JDBC_VERSION = "jdbc_version"
+  val CERTIFIED_JDBC_VERSION = "certified_jdbc_version"
+  val SFURL = "sfurl"
+  val OPERATION = "operation"
+  val QUERY = "query"
+  val QUERY_ID = "query_id"
+  val QUERY_STATUS = "query_status"
```
sfc-gh-mrui, line 257: To reviewer:
There are 2 actual fields name changes:
1. For QUERY_STATUS, "status" -> "query_status", this is only used in new telemetry message spark_query_status,  I believe it is no problem.
2. For QUERY_ID with released OOB message, "queryid" -> "query_id". This doesn't cause any change for customer. When we need to to query with filter on QUERY_ID in snowhouse, we can query both old and new values. So I think it should be fine.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-sshankar: let's not do this.

sfc-gh-sshankar (commented):
Are any field names actually changing in this PR ? Or are you just creating constants for everything ?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: I will update.

sfc-gh-mrui:
> Are any field names actually changing in this PR ? Or are you just creating constants for everything ?

The main purpose of this PR is creating constants for every telemetry field names. And make them to be defined in one place to avoid different messages to use different field name for the same purpose property.
The only actual field name change is that for OOB telemetry message the "queryid" is changed as "query_id".

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -243,80 +285,84 @@ object TelemetryConstValues {
...
+  val SPARK_CONNECTOR_VERSION = TelemetryFieldNames.SPARK_CONNECTOR_VERSION
   // The query operation: read vs write
-  val OPERATION: String = "operation"
+  val OPERATION = TelemetryFieldNames.OPERATION
   // Query statement
-  val QUERY: String = "query"
+  val QUERY = TelemetryFieldNames.QUERY
   // query ID if available
-  val QUERY_ID: String = "query_id"
+  val QUERY_ID = TelemetryFieldNames.QUERY_ID
```
sfc-gh-ema, line 294: These seem redundant: why not just directly use `TelemetryFieldNames.XXX`?

sfc-gh-mrui: Do you mean we don't need to use TelemetryQueryStatusFields?

My idea is that we'd better to classify the field names for each telemetry message separately in one place. It is much clear for maintenance and query. As more telemetry messages are added in the future, this will be more important.

sfc-gh-ema (approved):

sfc-gh-sshankar (approved):
Approval

PR #306, opened 2020-10-28
Add StressReadWriteSuite

sfc-gh-ema:


ClusterTest/src/main/scala/net/snowflake/spark/snowflake/testsuite/StressReadWriteSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+    * Reads the json file as configured by the system env variable defined in TestUtils.STRESS_TEST_SOURCES
+    * to get the tables to perform the read/write stress test on.
+    *
+    * The file should be in the format:
+    *
+    * [
+    *   {
+    *     "schema": "table1_schema",
```
sfc-gh-mrui, line 44: Nice to have: it should be better to add "database" for the json file to give the flexibility.

sfc-gh-mrui (commented):
It looks this PR is a part of the whole.
I think if the PR is mainly for testing, it should be fine to use one PR for all changes.

PR #307, opened 2020-10-30
Add StressTestSuite, and Refactor ClusterTest for Use in the Stress Test

sfc-gh-ema:
This PR, broadly speaking, makes a number of changes to ClusterTest, and adds a configurable StressTestSuite, which reads from a JSON file to configure which tables to perform read/write on. This is to enable running this our EMR stress-test cluster. 
In particular, the changes include (but are not limited to):

- Adding a TaskContext so that test suites can annotate timing and status with respect to subtasks within a suite (e.g., so StressTestSuite can keep track of the performance of each query, and whether it failed or succeeded).
- Refactoring code to separate suites from ResultBuilder: there are now different ResultBuilders depending on what behavior is needed, and ResultBuilders should be compatible with all suites, so that both Github Actions as well as the new stress-test environment can run all suites (i.e., the original ResultBuilder is now GithubActionsResultBuilder, and for the stress tests we have StressTestResultBuilder)
- Adding support for using a Vault instance to read login credentials.

There will be more to come, in particular, the shell script to instantiate the ClusterTest in the stress-test environment, as well as the JSON file specifying the test-data sources.


ClusterTest/src/main/scala/net/snowflake/spark/snowflake/BaseTestResultBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.snowflake.spark.snowflake
+
+abstract class BaseTestResultBuilder() {
```
sfc-gh-mrui, line 19: 1. "trait" vs "abstract class". Which should be better in this case.
2. Suppose we should put all new classes as private[snowflake]

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/ClusterTest.scala
```diff
@@ -30,6 +37,10 @@ object ClusterTest {
 
   // Driver function to run the test.
   def main(args: Array[String]): Unit = {
+    // If there are more than 2 arguments, we're running in the stress test environment
+    val envType: ClusterTest =
```
sfc-gh-mrui, line 41: It should be better to add a new argument to distinction the 2 kinds of run, instead of argument count.
lets say, the 3rd argument can be "githubtest" or "stresstest".

sfc-gh-mrui: If we do this, please update the script in ClusterTest.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/StressTestResult.scala
```diff
@@ -0,0 +1,108 @@
...
+           | reason String )
+           |""".stripMargin)
+    }
+
+    // Get the next runID
+    val res = DefaultJDBCWrapper
+      .executeQueryInterruptibly(connection, s"select ${TestUtils.STRESS_TEST_SEQ_NAME}.nextVal")
+    res.next()
+
+    val runId = res.getInt(1)
```
sfc-gh-mrui, line 57: What's the default return datatype for SEQUENCER? If it is Long, may we need to use res.getLong(1).

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/StressTestResult.scala
```diff
@@ -0,0 +1,108 @@
...
+    val runTableName = TestUtils.STRESS_TEST_RUN_TABLE_PREFIX + runId
+
+    // A table with this name should not exist; if it does, something is wrong.
+    // We need to terminate this early
+    if (DefaultJDBCWrapper.tableExists(connection, runTableName)) {
```
sfc-gh-mrui, line 62: I think it is not necessary to create a table for each test suite run. What about to use two tables: 
1. TestUtils.STRESS_TEST_RESULT_TABLE is used to record the total time for one test suite run.
2. TestUtils.STRESS_TEST_RESULT_TABLE_DETAILS (new table) is used to record the test test for each test case.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/TaskContext.scala
```diff
@@ -0,0 +1,8 @@
+package net.snowflake.spark.snowflake
+
+private[snowflake] case class TaskContext(var testName: String) {
```
sfc-gh-mrui, line 3: What about using TestContext or TestSuiteContext to distinguish with Spark TaskContext?

sfc-gh-mrui (changes requested):
Thanks for the great work. I have added some comments.
In general, please also work on below after this PR is merged.
1. Update https://github.com/snowflakedb/spark-snowflake/blob/master/README-TEST.md
2. Update wiki page https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/1016168969/Spark+Connector+Test+Framework
Of course, you can write details at one place and refer to it from another.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/BaseTestResultBuilder.scala
sfc-gh-ema: I've added the package-private modifiers to these classes.

I don't think there's a clear reason yet to use trait vs. class, since we don't expect `BaseTestResultBuilder` to be mixed-in with other types. Classes are slightly more efficient so I'll keep using them here.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/StressTestResult.scala
sfc-gh-ema: Changed.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/StressTestResult.scala
sfc-gh-ema: I think that makes sense. I changed it.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/ClusterTest.scala
sfc-gh-ema: Done.

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/TaskContext.scala
sfc-gh-ema: Changed.

sfc-gh-ema:
> Thanks for the great work. I have added some comments.
> In general, please also work on below after this PR is merged.
> 
> 1. Update https://github.com/snowflakedb/spark-snowflake/blob/master/README-TEST.md
> 2. Update wiki page https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/1016168969/Spark+Connector+Test+Framework
>    Of course, you can write details at one place and refer to it from another.

Done! PTAL.

ClusterTest/project/assembly.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
```
sfc-gh-mrui, line 1: For my learning purpose, what's this plug-in for?

ClusterTest/run_stress_test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+export TEST_COMPILE_SCALA_VERSION=2.12
+export TEST_SPARK_CONNECTOR_VERSION=2.8.2
+export TEST_SPARK_VERSION=3.0
+export TEST_CLUSTERTEST_VERSION=1.0
+
+# Test Revision ID
+export TEST_REVISION_ID=1
```
sfc-gh-mrui, line 9: Can you add comment to explain TEST_REVISION_ID and when it should be bump up.

build.sbt
```diff
@@ -68,6 +68,11 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
       // "org.apache.spark" %% "spark-hive" % testSparkVersion % "provided, test"
     ),
 
+    assemblyMergeStrategy in assembly := {
```
sfc-gh-mrui, line 71: For my learning purpose, what's assemblyMergeStrategy for?

README-TEST.md
```diff
@@ -196,4 +196,39 @@ To better understand why this 227 MB file requires around at least 600MB memory
...
+  },
+  {
+    "database": "testdb_spark",
+    "schema": "tpch_sf1",
+    "table": "lineitem"
+  }
+]
+```
+
+Each item in the array denotes a table that `StressReadWriteSuite` will read from; each table will be formulated as a subtask in `StressReadWriteSuite`. When this JSON is read, each item in the array will run as a subtask in the overall `StressReadWriteSuite`.
```
sfc-gh-mrui, line 234: If I understand correctly, if we want to test one more table in the future, what we need to do is to add one entry in ClusterTest/test_sources.json. Correct? If anything else we need to do, could you please describe here?

sfc-gh-mrui (commented):
In general, it is good. Have some comments.

build.sbt
sfc-gh-ema: This resolves dependencies when there are conflicts in versions during the build process. This can occur when the same dependency is cached in different places; this strategy AFAIK simply discards one of them and uses the other.

README-TEST.md
sfc-gh-ema: That's right, we just need to edit `test_sources.json` with the appropriate entry.

ClusterTest/project/assembly.sbt
sfc-gh-ema: This is for running `sbt assembly`. For convenience purposes when running `spark-submit` in our stress infrastructure, we bundle all of the dependencies into fat JARs; assembly builds the JARs with all dependencies included.

ClusterTest/run_stress_test.sh
sfc-gh-ema: Done!

ClusterTest/project/assembly.sbt
sfc-gh-mrui: Thanks!

build.sbt
sfc-gh-mrui: It will be great if you can add comment for this. It doesn't block your current PR. You can do in next PR.

sfc-gh-mrui (approved):
LGTM

build.sbt
sfc-gh-ema: Done.

PR #310, opened 2020-11-13
SNOW-222104 Fix a writing issue when table name includes Database or Schema

sfc-gh-mrui:
Issue description
============
In the PR 282, all special characters are replaced as 'X' when generating a temp stage table name for writing to snowflake database. But unfortunately, some users have the table name to include database and schema name. If below 4 conditions are satisfied in the same time, the write job will fail.
1. Write data frame to a table with OVERWRITE and (usestagingtable=on truncate_table=off (they are default)).
2. table name includes database name and schema name.
3. sfSchema is configured to a different schema
4. The user has privilege to create stage but doesn't have privilege to create table on sfSchema

Fix
====
There is a basic assumption that the table name must be a valid table name for snowflake,
otherwise create table will fail anyway. The safest way to generate a staging table name is
to add a random postfix to the original table name. And the random postfix only include
common characters [0-9a-zA-Z ]. The database name and schema name are in the leading part,
so they are not affected. For example,
My_DB.My_Schema.MY_TABLE        ==> My_DB.My_Schema.MY_TABLE_staging_XXXXX
"My.DB"."My.Schema"."MY.TABLE"  ==> "My.DB"."My.Schema"."MY.TABLE_staging_XXXXX"

Be careful to handle below special cases:
1. Table name may be quoted or not.
2. There may be white characters in the end.

Below are the unit test table names and the generated stage table names:

test_table  --->  test_table_staging_868568613
test_table!@#$%^&^  --->  test_table!@#$%^&^_staging_1219750424
My_Schema.Table_1  --->  My_Schema.Table_1_staging_1198998292
My_DB.My_Schema.Table_1  --->  My_DB.My_Schema.Table_1_staging_2010887278
"My.Schema".Table_1  --->  "My.Schema".Table_1_staging_1773042307
"My.DB"."My.Schema".Table_1  --->  "My.DB"."My.Schema".Table_1_staging_1457942305
"test_table_.'!@#$%^&* 2611611920364743726"   --->  "test_table_.'!@#$%^&* 2611611920364743726_staging_638657005"
"Table.1"  --->  "Table.1_staging_1242308192"
"My.Schema"."Table.1"  --->  "My.Schema"."Table.1_staging_2049518994"
"My.Schema"."Table.1"   --->  "My.Schema"."Table.1_staging_900935185"
"My.DB"."My.Schema"."Table.1"  --->  "My.DB"."My.Schema"."Table.1_staging_1346640764"
My_Schema."Table.1"  --->  My_Schema."Table.1_staging_451508044"
My_DB."My.Schema"."Table.1"  --->  My_DB."My.Schema"."Table.1_staging_2136103915"
"My.DB".My_Schema."Table.1"  --->  "My.DB".My_Schema."Table.1_staging_2119220707"
My_Schema."Table""1"  --->  My_Schema."Table""1_staging_1395092498"
"My""Schema"."Table""""2"  --->  "My""Schema"."Table""""2_staging_1874852802"
My_DB."My.Schema"."Table""1"  --->  My_DB."My.Schema"."Table""1_staging_1991416482"
My_DB."My""Schema"."Table""""2"  --->  My_DB."My""Schema"."Table""""2_staging_785611724"
My_DB."My""Schema"."Table""2"""  --->  My_DB."My""Schema"."Table""2""_staging_66100706"

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -130,9 +130,9 @@ object Parameters {
     "use_exponential_backoff"
   )
   // Internal option to remove (") in stage table name.
-  // It may be removed without any notice in any time.
-  val PARAM_INTERNAL_REMOVE_QUOTE_FOR_STAGE_TABLE_NAME: String = knownParam(
-    "internal_remove_quote_for_stage_table_name"
+  // This option may be removed without any notice in any time.
+  val PARAM_INTERNAL_STAGING_TABLE_NAME_REMOVE_QUOTES_ONLY: String = knownParam(
+    "internal_staging_table_name_remove_quotes_only"
   )
 
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
@@ -582,8 +582,8 @@ object Parameters {
     def useExponentialBackoff: Boolean = {
       isTrue(parameters.getOrElse(PARAM_USE_EXPONENTIAL_BACKOFF, "false"))
     }
-    def removeQuoteForStageTableName: Boolean = {
-      isTrue(parameters.getOrElse(PARAM_INTERNAL_REMOVE_QUOTE_FOR_STAGE_TABLE_NAME, "false"))
+    def stagingTableNameRemoveQuotesOnly: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_INTERNAL_STAGING_TABLE_NAME_REMOVE_QUOTES_ONLY, "false"))
     }
 
     /**
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -129,6 +129,11 @@ object Parameters {
   val PARAM_USE_EXPONENTIAL_BACKOFF: String = knownParam(
     "use_exponential_backoff"
   )
+  // Internal option to remove (") in stage table name.
+  // It may be removed without any notice in any time.
+  val PARAM_INTERNAL_REMOVE_QUOTE_FOR_STAGE_TABLE_NAME: String = knownParam(
+    "internal_remove_quote_for_stage_table_name"
+  )
 
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
   val MIN_S3_MAX_FILE_SIZE = 1000000
@@ -576,6 +581,9 @@ object Parameters {
     }
     def useExponentialBackoff: Boolean = {
       isTrue(parameters.getOrElse(PARAM_USE_EXPONENTIAL_BACKOFF, "false"))
+    }
+    def removeQuoteForStageTableName: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_INTERNAL_REMOVE_QUOTE_FOR_STAGE_TABLE_NAME, "false"))
     }
 
     /**
--- a/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
@@ -323,16 +323,16 @@ private[io] object StageWriter {
     def genTableName():String =
       s"spark_stage_table_${System.currentTimeMillis()}_${Math.abs(Random.nextInt)}"
 
-    val tableNameQuoted = "(.*)\"([^\"]*)\"(\\s*)".r
-    val tableNameNotQuoted = "(.*)\\.([^.]*)".r
-
+    val pattern = "(.*\\.)?((\".+\")|([^\"]+))(\\s*)".r
     tableName match {
-      case tableNameQuoted(databaseSchemaPart, _, _) =>
-        s"$databaseSchemaPart${genTableName()}"
-      case tableNameNotQuoted(databaseSchemaPart, _) =>
-        s"$databaseSchemaPart.${genTableName()}"
-      case _ =>
-        s"${genTableName()}"
+      case pattern(dbAndSchema, _, _, _, _) => {
+        s"${if (dbAndSchema == null) "" else dbAndSchema}${genTableName()}"
+      }
+      case _ => {
+        // For example, 'table_1"' or '"table_1' is an illegal name.
+        log.warn(s"The table name is illegal: {{$tableName}}")
+        genTableName()
+      }
     }
   }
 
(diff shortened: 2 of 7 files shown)


src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -319,6 +319,23 @@ private[io] object StageWriter {
...
+
+    val tableNameQuoted = "(.*)\"([^\"]*)\"(\\s*)".r
+    val tableNameNotQuoted = "(.*)\\.([^.]*)".r
+
+    tableName match {
+      case tableNameQuoted(databaseSchemaPart, _, _) =>
+        s"$databaseSchemaPart${genTableName()}"
+      case tableNameNotQuoted(databaseSchemaPart, _) =>
+        s"$databaseSchemaPart.${genTableName()}"
+      case _ =>
```
sfc-gh-bli, line 334: how about this name `DB."schema".TABLE`?

sfc-gh-mrui: The table name will be generated as `DB."schema".spark_stage_table_XXX_YYY`

Unit test covers below:
My_DB."My.Schema"."Table.1" ---> My_DB."My.Schema".spark_stage_table_1605311023640_940705254

sfc-gh-bli: `aa.`?

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -319,6 +319,23 @@ private[io] object StageWriter {
     }
   }
 
+  private[snowflake] def getStageTableName(tableName: String): String = {
+    def genTableName():String =
+      s"spark_stage_table_${System.currentTimeMillis()}_${Math.abs(Random.nextInt)}"
+
+    val tableNameQuoted = "(.*)\"([^\"]*)\"(\\s*)".r
+    val tableNameNotQuoted = "(.*)\\.([^.]*)".r
```
sfc-gh-bli, line 327: Try this
```scala
"(.*)((\"[^\"]+\")|([^.]+))(\\s*)".r
```

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -319,6 +319,23 @@ private[io] object StageWriter {
     }
   }
 
+  private[snowflake] def getStageTableName(tableName: String): String = {
+    def genTableName():String =
+      s"spark_stage_table_${System.currentTimeMillis()}_${Math.abs(Random.nextInt)}"
+
```
sfc-gh-bli, line 325:
```scala
val pattern = "(.*\\.)?((\".+\")|([^\"]+))(\\s*)".r
tableName match {
    case pattern(dbAndSchema, _, _, _, _) => println(dbAndSchema)
    case _ => throw new Exception()
}
```

sfc-gh-bli: try this. It can match `"aa".bb."cc""dd"`

sfc-gh-mrui: You are appreciated so much! It works.
I also add more unit test for table name including QUOTE character.

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -333,10 +345,7 @@ private[io] object StageWriter {
                                            fileUploadResults: List[FileUploadResult])
   : Unit = {
     val table = params.table.get
-    val tempTable =
```
sfc-gh-ema, line 336: Can you add a parameter to disable this new behavior? My fear is that some customers are using the old staging table name scheme.

sfc-gh-mrui (commented):
@sfc-gh-sshankar @sfc-gh-bli @sfc-gh-ema I will add an internal option to disable the name change. Could you please take a look at the analysis about reason? Any comments are welcome!

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: After more consideration, the regression about item-4 should not exist. If the schema name is MY_SCHEMA, the table rename or swap when committing the WRITE will fail.

sfc-gh-bli (approved):
LGTM

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-sshankar: I don't understand the concern here. We are creating this temp table, how can customers be using it ?

sfc-gh-mrui: @sfc-gh-sshankar The staging table only exists when writing a data frame to snowflake. But the table is a normal table (not a snowflake TEMP table). So it could be seen by others users when it exists.
I think it is extremely low possibility that the customers depends on the staging table name. In theory, it still is possible. (The example I can imagine is that the customer may clean up some unused staging tables like "%_staging_%", In most cases, if the writing job fails, the staging table will be removed automatically.)

@sfc-gh-ema Please supplement if you have more specific examples.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -129,6 +129,11 @@ object Parameters {
   val PARAM_USE_EXPONENTIAL_BACKOFF: String = knownParam(
     "use_exponential_backoff"
   )
+  // Internal option to remove (") in stage table name.
+  // It may be removed without any notice in any time.
+  val PARAM_INTERNAL_REMOVE_QUOTE_FOR_STAGE_TABLE_NAME: String = knownParam(
```
sfc-gh-sshankar, line 134: rename to PARAM_INTERNAL_STAGING_TABLE_NAME_REMOVE_QUOTES_ONLY

sfc-gh-mrui: thanks. Update it.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-sshankar: Write this as follows: 
if (param) {
old behavior
} else {
new behavior
}

sfc-gh-sshankar (commented):
Done, minor rewrites required. In geneal, it's clearer if you can make the behavior of the parameter clear int he code.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: refactor it.

sfc-gh-sshankar (approved):
Approval

PR #316, opened 2020-12-03
Code Review for SC 2.8.3 for spark 2.3 (This PR is for code-review only, it is NOT merged)

sfc-gh-mrui:


sfc-gh-bli (approved):

.github/workflows/IntegrationTest.yml
```diff
@@ -2,18 +2,18 @@ name: Spark Connector Integration Test
...
 jobs:
   test:
     name: IT
     runs-on: ubuntu-latest
     strategy:
      matrix:
-       scala_version: [ '2.12.11' ]
-       spark_version: [ '3.0.0' ]
+       scala_version: [ '2.11.12' ]
+       spark_version: [ '2.3.0' ]
```
sfc-gh-ema, line 16: 2.3.0 over here, 2.3.4 above. Is this intentional?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/StringStatement.scala
```diff
@@ -11,6 +11,7 @@ import org.apache.spark.sql.catalyst.expressions.{
   Lower,
   StringLPad,
   StringRPad,
+  StringReverse,
```
sfc-gh-ema, line 14: Does this mean we actually support more pushdowns in 2.3 than for 3.0?

sfc-gh-ema (approved):
Approved with some minor comments.

.github/workflows/IntegrationTest.yml
sfc-gh-mrui: Yes. It is intentional. The IT test is run for spark 2.3.0/2.4.0/3.0.0. We didn't actually run for all released versions.
When I work on Cluster Test, I used some new versions: 2.3.4/2.4.6/3.0.0.

Since the SPARK should keep the interface consistent. I believe it is not a problem. And we didn't get any customer to report this kind of compatibility issue.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/StringStatement.scala
sfc-gh-mrui: StringReverse is not supported from Spark 2.4, so SC needs to remove it accordingly.
So, the pushdown support is the same actually.

PR #317, opened 2020-12-04
SNOW-187770 Support to write large partition with less memory for AWS

sfc-gh-mrui:
SNOW-187770
If the data size is bigger than the chunk size, AWS multiple parts upload
API will be used to upload with less memory.
If the data size is smaller than it, current normal upload API is still used.
NOTE: this change only affects writing to AWS with internal stage.


ClusterTest/src/main/scala/net/snowflake/spark/snowflake/testsuite/LowMemoryStressSuite.scala
```diff
@@ -76,13 +76,25 @@ class LowMemoryStressSuite extends ClusterTestSuiteBase {
...
+    // Write to snowflake
+    df.write
+      .format(TestUtils.SNOWFLAKE_NAME)
+      .options(TestUtils.sfOptionsNoTable)
+      .option("dbtable", test_big_partition)
+      .mode(SaveMode.Overwrite)
+      .save()
+
+    log.info(
+      s"""Finished the first multi-part upload test ======================""".stripMargin)
```
sfc-gh-bli, line 88: I prefer not including `=====` in the log message

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -1496,9 +1498,252 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
...
+          .option(Parameters.PARAM_UPLOAD_CHUNK_SIZE_IN_MB, "5")
+          .option("dbtable", test_table_write)
+          .mode(SaveMode.Overwrite)
+          .save()
+      })
+      TestHook.disableTestHook()
+    }
+  }
+
+  ignore("perf test for AWS multiple parts upload API") {
```
sfc-gh-bli, line 1585: why we ignore this test?

sfc-gh-bli (commented):
Any performance change in this new method?

ClusterTest/src/main/scala/net/snowflake/spark/snowflake/testsuite/LowMemoryStressSuite.scala
sfc-gh-mrui: "====" is included just because it is there when Zihan developed this case. I just get it back. I agree it is not necessary. I will remove it.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
sfc-gh-mrui: This test case is used to evaluate the performance with old API and new API with different buffer size. It's more like a manual test and it needs a long time time to run. So it is not necessary to enable it.

sfc-gh-mrui: And this test indicates there is no perf difference with old API and new API with different buffer size.

sfc-gh-mrui:
> Any performance change in this new method?

No change for the performane.
The performance with old API and new API is the same. And the new API with different buffer size is also near the same.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -1860,3 +1880,184 @@ object StorageUtils {
   }
 }
 
+private[io] class S3UploadOutputStream(s3Client: AmazonS3Client,
```
sfc-gh-ema, line 1883: Quick question, if the job fails part way and some of the parts are uploaded already, is there any impact?

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -1860,3 +1880,184 @@ object StorageUtils {
...
+    totalDataSize += dataSizeInBuffer
+    dataSizeInBuffer = 0
+    partId += 1
+    byteArrayOutputStream.reset()
+    // byteArrayOutputStream = new ByteArrayOutputStream(bufferSize + 8)
+
+    val usedTime = System.currentTimeMillis() - start
+    totalUploadTime += usedTime
+
+    if (partId == 3) {
```
sfc-gh-ema, line 1993: ?

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: Below description from AWS document addresses your question. The updated parts reside in the stage and AWS suggest to configure some strategy to clean up it. This is not a problem for our use case because the internal stage is temp stage, so it will be cleaned up in the end. 

=====================
After you initiate a multipart upload, you begin uploading parts. Amazon S3 stores these parts, but it creates the object from the parts only after you upload all of them and send a successful request to complete the multipart upload (you should verify that your request to complete multipart upload is successful). Upon receiving the complete multipart upload request, Amazon S3 assembles the parts and creates an object.

If you don't send the complete multipart upload request successfully, Amazon S3 will not assemble the parts and will not create any object. Therefore, the parts remain in Amazon S3 and you pay for the parts that are stored in Amazon S3. As a best practice, we recommend you configure a lifecycle rule (using the AbortIncompleteMultipartUpload action) to minimize your storage costs.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: I guess your question is why "if (partId == 3)" is used. correct?
It only injects an exception when uploading the 2nd part.

sfc-gh-mrui: There should be no difference to inject an exception with other parts. I thought it may be more common to hit upload error on non-first part.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html#mpu-stop-incomplete-mpu-lifecycle-config

sfc-gh-sshankar (approved):
Approval, assuming clean cherry pick

sfc-gh-mrui:
> Approval, assuming clean cherry pick

ACK. The cherry pick is clean.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-ema: I guess my confusion is why this piece of code that's for testing on the same code path as production code?

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: Thanks for the clarification.
This is used to simulate an upload error by raising an exception if the flag is ON.
The flag is enabled only if the some test. For all the others, there is an extra IF check. 
But, the perf side-effects for production code can be ignored.

sfc-gh-ema: Ah I see, I guess it just still seems a bit weird to have this on the same code path as regular production. Is there any other way to raise an exception reliably for your test?

sfc-gh-mrui: In C/C++, we can use conditional compilation such as "#ifdef ... #endif".
But, I didn't find the equivalent functionality in Java/Scala. If you or bing have any idea about this, please let me know. I can refactor it.

sfc-gh-ema: How about this:

Close the stream from your test after uploading two parts. If your stream is closed, trying to upload more parts should raise an exception.

sfc-gh-mrui: The test case is writing a data frame to a snowflake table. It is difficult because
1. It needs some hacking to let the test to get the stream object.
2. The test is difficult to catch the timing to close the stream.

sfc-gh-ema (approved):

PR #318, opened 2020-12-08
SNOW-239416 Disable one test case for SC regress test.

sfc-gh-mrui:
The test case of "write table with different schema" in
TruncateTableSuite needs special test account configuration to run it.
But the account configuration is not done for preprod3/QA account,
so disable this test case for the released SC back-compatibility
regress test.


src/it/scala/net/snowflake/spark/snowflake/TruncateTableSuite.scala
```diff
@@ -384,7 +384,10 @@ class TruncateTableSuite extends IntegrationSuiteBase {
   // 2. Configure truncate_table = on and usestagingtable=off can workaround this issue.
   test("write table with different schema") {
     val accountName = System.getenv(SNOWFLAKE_TEST_ACCOUNT)
```
sfc-gh-mrui, line 386: To reviewers:
This fix only skips to run the test case for released SC back-compatibility regress test.
The test is still included in IT test in Github Action and Travis.

sfc-gh-ema (approved):

sfc-gh-bli (approved):

PR #322, opened 2020-12-11
SNOW-241656 Enhance test function testPushdown().

sfc-gh-mrui:
In spark connector pushdown test,  testPushdown() is commonly used. It will
1. Read the DataFrame from snowflake.
2. Compare the result to be equal to the expected result.
3. The executed query text must match the expected query text.

The read result must the be the same even if the pushdown is disabled. So, add below 2 extra steps for this test function.
1. Disable pushdown and read the DataFrame again.
2. Compare the read result to be equal to the expected result. (Don’t check executed the query text).

This enhancement has found 3 test cases failure. After investigation, they are not actual Spark connector issues. So disable the extra steps for them. Explain them in file changes.

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/IntegrationSuiteBase.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/IntegrationSuiteBase.scala
@@ -18,6 +18,7 @@
 package net.snowflake.spark.snowflake
 
 import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
+import net.snowflake.spark.snowflake.pushdowns.SnowflakeStrategy
 import org.apache.spark.sql._
 import org.apache.spark.sql.types.StructType
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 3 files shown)


src/it/scala/net/snowflake/spark/snowflake/PushdownEnhancement01.scala
```diff
@@ -105,7 +105,8 @@ class PushdownEnhancement01 extends IntegrationSuiteBase {
          |FROM ( SELECT * FROM ( $test_table_length ) AS
          |"SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0" """.stripMargin,
       result,
-      expectedResult
+      expectedResult,
+      testPushdownOff = false
```
sfc-gh-mrui, line 109: To Reviewers:
If use_copy_unload = true, the test fails. This difference comes from the limitation for COPY UNLOAD NULL value processing. It can be ignored.

1. When reading with COPY UNLOAD, empty string will be read as NULL. (With Arrow, it is ready as "").
2. This test case test: length(""), length("abc"), length(null).
3. The expected result is (0, 3, NULL).
4. With COPY UNLOAD, the data is read as (null, "abc", null). So the final result is (null, 3, null).

sfc-gh-bli (approved):

sfc-gh-ema (approved):

src/it/scala/net/snowflake/spark/snowflake/IntegrationSuiteBase.scala
```diff
@@ -46,11 +47,11 @@ trait IntegrationSuiteBase
     * Verify that the pushdown was done by looking at the generated SQL,
     * and check the results are as expected
     */
-  def testPushdown(reference: String,
-                   result: DataFrame,
-                   expectedAnswer: Seq[Row],
-                   bypass: Boolean = false,
-                   printSqlText: Boolean = false): Unit = {
+  def testPushdownBasic(reference: String,
```
sfc-gh-sshankar, line 50: Should this be private

sfc-gh-mrui: updated it.

PR #324, opened 2021-01-12
SNOW-250653 Quote Json field name when writing with Json format.

sfc-gh-mrui:


src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -740,13 +740,25 @@ private[io] object StageWriter {
         case SupportedFormat.JSON =>
           if (list.isEmpty || list.get.isEmpty) {
             val names = schema.fields
-              .map(x => "parse_json($1):".concat(x.name))
+              .map(x => "parse_json($1):".concat(
+                if (params.quoteJsonFieldName) {
+                  "\"" + x.name + "\""
```
sfc-gh-bli, line 745: escape?

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: I think you mean we need to escape (") in the x.name. Correct?
It is separate with this Jira. Current we don't support the column to have (") character when writing a dataframe with OVERWRITE. Below test fails.

  test("test column name has quotation") {
    val df = sparkSession.range(10).toDF(s"""for"some""")
    df.write
      .options(thisConnectorOptionsNoTable)
      .option("dbtable", test_table_write)
      .mode(SaveMode.Overwrite)
      .save()
  }

It fails when create table. But no customer reports this, I will file a Jira for it. But I don't think we need to fix it now.

sfc-gh-mrui: @sfc-gh-bli @sfc-gh-ema @sfc-gh-sshankar Separate Jira https://snowflakecomputing.atlassian.net/browse/SNOW-259724 is filed for this issue. FYI.

sfc-gh-bli (approved):
LGTM

sfc-gh-sshankar (approved):
Approval

PR #326, opened 2021-01-26
SNOW-262080 use fully qualified name when detecting table existence

sfc-gh-mrui:
Issue description
============
When writing to a table, SC detects the target table existence with `desc table <tableName>`. Suppose the table name is `t1` and `sfschema` is set as `schema1`, if there is no `schema1.t1` table but `public.t1` exists, the `desc table t1` will return the table of `public.t1`. This is described in snowflake document: https://docs.snowflake.com/en/sql-reference/name-resolution.html#resolution-when-schema-omitted-double-dot-notation . But it is not SC's expectation.

Fix proposal
=========
SC uses the full qualified table name to detect the table existence. This needs to be handled very carefully. Make sure below cases are considered:
1. table name may be quoted or not.
2. table name may include special characters.
3. table name may includes schema name only, database name only or both database and schema name.
4. sfSchema and sfDatabase may include special characters or be quoted.

Implementation
============
1. Introduce an internal option `internal_use_full_qualified_name` to disable the fix if necessary.
2. Implement a private function `Utils.splitNameAs2Parts(name)` to split table name from the full name. For example, it splits `database1.schema1.table1` as (`database1.schema1`, `table1`); it splits `"database1"."schema1"` as (`"database1"`, `"schema1"`).
3. Implement a private function `Utils.getFullQualifiedName(name, params)` to get the full qualified name. If database name or schema name are not included in the original name, use sfdatabase or sfschema. For example,
- sfdatabase=test_db, sfschema=test_schema, table name is table1, the full name is `"TEST_DB"."TEST_SCHEMA".table1`.
- sfdatabase=test_db, sfschema=test_schema, table name is schema1.table1, the full name is `"TEST_DB".schema1.table1`.
For more cases, refer to test of `Utils.getFullQualifiedName()` in `UtilsSuite.scala`
4. If `internal_use_full_qualified_name` is true, use the full qualified name when detecting table existence in `StageWriter`

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -153,11 +153,11 @@ object Parameters {
   val PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME: String = knownParam(
     "internal_quote_json_field_name"
   )
-  // Internal option to use full qualified table name
+  // Internal option to use fully qualified table name
   // when detecting the table existence.
   // This option may be removed without any notice in any time.
-  val PARAM_INTERNAL_USE_FULL_QUALIFIED_NAME: String = knownParam(
-    "internal_use_full_qualified_name"
+  val PARAM_INTERNAL_USE_FULLY_QUALIFIED_NAME: String = knownParam(
+    "internal_use_fully_qualified_name"
   )
 
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
@@ -615,8 +615,8 @@ object Parameters {
     def quoteJsonFieldName: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME, "true"))
     }
-    def useFullQualifiedName: Boolean = {
-      isTrue(parameters.getOrElse(PARAM_INTERNAL_USE_FULL_QUALIFIED_NAME, "true"))
+    def useFullyQualifiedName: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_INTERNAL_USE_FULLY_QUALIFIED_NAME, "true"))
     }
 
     /**
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -153,6 +153,12 @@ object Parameters {
   val PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME: String = knownParam(
     "internal_quote_json_field_name"
   )
+  // Internal option to use full qualified table name
+  // when detecting the table existence.
+  // This option may be removed without any notice in any time.
+  val PARAM_INTERNAL_USE_FULL_QUALIFIED_NAME: String = knownParam(
+    "internal_use_full_qualified_name"
+  )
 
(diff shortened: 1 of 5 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -153,6 +153,12 @@ object Parameters {
   val PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME: String = knownParam(
     "internal_quote_json_field_name"
   )
+  // Internal option to use full qualified table name
+  // when detecting the table existence.
+  // This option may be removed without any notice in any time.
+  val PARAM_INTERNAL_USE_FULL_QUALIFIED_NAME: String = knownParam(
+    "internal_use_full_qualified_name"
```
sfc-gh-sshankar, line 160: internal\_use\_fully\_qualified\_name

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -712,4 +712,51 @@ object Utils {
...
+      if (objectName.indexOf(".") > -1) {
+        (objectName.substring(0, objectName.lastIndexOf(".")),
+          objectName.substring(objectName.lastIndexOf(".") + 1))
+      } else {
+        ("", objectName)
+      }
+    }
+  }
+
+  private[snowflake] def getFullQualifiedName(originalName: String, params: MergedParameters): String = {
```
sfc-gh-ema, line 736: Similar to Srinath's comment above: `getFullyQualifiedName`.

sfc-gh-ema (approved):

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -281,16 +281,21 @@ private[io] object StageWriter {
     val writeTableState = new WriteTableState(conn)
 
     try {
+      val fullyQualifiedName = if (params.useFullyQualifiedName) {
```
sfc-gh-sshankar, line 284: This is bad naming. If the param is false, fullyQualifiedName is actually not full qualified.
Just do
```
val tableName = if (params.useFullyQualifiedName) { 
  Utils.getFullyQualifiedName(params.table.get.name, params)
} else {
  params.table.get.name
}
```

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -362,9 +367,14 @@ private[io] object StageWriter {
       }
 
     try {
+      val fullyQualifiedName = if (params.useFullyQualifiedName) {
```
sfc-gh-sshankar, line 370: Same see above

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -712,4 +712,53 @@ object Utils {
         println(s"Fail to print result set: ${th.getMessage}")
     }
   }
+
+  private[snowflake] def splitNameAs2Parts(objectName: String): (Option[String], String) = {
+    if (objectName.endsWith("\"")) {
```
sfc-gh-sshankar, line 717: How do you handle quotes within names ?

sfc-gh-mrui:
Use a lower risk proposal to fix the issue. Close this PR.
https://github.com/snowflakedb/spark-snowflake/pull/327

PR #327, opened 2021-02-01
SNOW-262080 check table existence in current schema only

sfc-gh-mrui:
Issue description
============
When writing to a table, SC detects the target table existence with desc table <tableName>. Suppose the table name is t1 and sfschema is set as schema1, if there is no schema1.t1 table but public.t1 exists, the desc table t1 will return the table of public.t1. This is described in snowflake document: https://docs.snowflake.com/en/sql-reference/name-resolution.html#resolution-when-schema-omitted-double-dot-notation . But it is not SC's expectation.

Fix proposal
=========
When writing to a snowflake table, SC checks the table existence in current schema only. To play it safe, a new JDBC connection is created and set the search path as current schema only. For example, 
```
conn.createStatement().execute("alter session set search_path='$current'")
```
In addition, add a new internal option 'internal_check_table_existence_in_current_schema_only' to disable this behavior if necessary. Its default is true.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
@@ -111,7 +111,7 @@ class DefaultSource(jdbcWrapper: JDBCWrapper)
 
     def tableExists: Boolean =
       if (params.checkTableExistenceInCurrentSchemaOnly) {
-        DefaultJDBCWrapper.tableExistsInCurrentSchema(params, table.toString)
+        jdbcWrapper.tableExistsInCurrentSchema(params, table.toString)
       } else {
         val conn = jdbcWrapper.getConnector(params)
         try {
--- a/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
@@ -109,14 +109,17 @@ class DefaultSource(jdbcWrapper: JDBCWrapper)
       )
     }
 
-    def tableExists: Boolean = {
-      val conn = jdbcWrapper.getConnector(params)
-      try {
-        jdbcWrapper.tableExists(conn, table.toString)
-      } finally {
-        conn.close()
+    def tableExists: Boolean =
+      if (params.checkTableExistenceInCurrentSchemaOnly) {
+        DefaultJDBCWrapper.tableExistsInCurrentSchema(params, table.toString)
+      } else {
+        val conn = jdbcWrapper.getConnector(params)
+        try {
+          jdbcWrapper.tableExists(conn, table.toString)
+        } finally {
+          conn.close()
+        }
       }
-    }
 
     val (doSave, dropExisting) = saveMode match {
       case SaveMode.Append => (true, false)
--- a/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
@@ -109,17 +109,7 @@ class DefaultSource(jdbcWrapper: JDBCWrapper)
       )
     }
 
-    def tableExists: Boolean =
-      if (params.checkTableExistenceInCurrentSchemaOnly) {
-        jdbcWrapper.tableExistsInCurrentSchema(params, table.toString)
-      } else {
-        val conn = jdbcWrapper.getConnector(params)
-        try {
-          jdbcWrapper.tableExists(conn, table.toString)
-        } finally {
-          conn.close()
-        }
-      }
+    def tableExists: Boolean = jdbcWrapper.tableExists(params, table.toString)
 
     val (doSave, dropExisting) = saveMode match {
       case SaveMode.Append => (true, false)
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -153,6 +153,15 @@ object Parameters {
   val PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME: String = knownParam(
     "internal_quote_json_field_name"
   )
+  // Internal option to check table existence in current schema only.
+  // By default, if customer specifies an unqualified table name
+  // (without schema or database name), snowflake checks table existence
+  // through a search path. For example, it checks current schema and PUBLIC
+  // schema.
+  // This option may be removed without any notice in any time.
+  val PARAM_INTERNAL_CHECK_TABLE_EXISTENCE_IN_CURRENT_SCHEMA_ONLY: String = knownParam(
+    "internal_check_table_existence_in_current_schema_only"
+  )
 
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
   val MIN_S3_MAX_FILE_SIZE = 1000000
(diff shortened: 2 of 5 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -287,10 +287,30 @@ private[snowflake] class JDBCWrapper {
...
   def tableExists(conn: Connection, table: String): Boolean =
     conn.tableExists(table)
 
+  /**
+    * Returns true if the table already exists in the current schema.
+    */
+  private[snowflake] def tableExistsInCurrentSchema(params: MergedParameters, table: String): Boolean = {
+    val conn = getConnector(params)
+    try {
+      conn.createStatement().execute("alter session set search_path='$current'")
```
sfc-gh-bli, line 307: do we want to change it back the default? if not, why don't we set this session parameter when login.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -281,16 +281,19 @@ private[io] object StageWriter {
...
+      val tableExists = if (params.checkTableExistenceInCurrentSchemaOnly){
+        DefaultJDBCWrapper.tableExistsInCurrentSchema(params, tableName)
+      } else {
+        DefaultJDBCWrapper.tableExists(conn, tableName)
+      }
       // Drop table only if necessary.
-      if (saveMode == SaveMode.Overwrite &&
-        DefaultJDBCWrapper.tableExists(conn, tableName) &&
-        !params.truncateTable )
+      if (saveMode == SaveMode.Overwrite && tableExists && !params.truncateTable)
```
sfc-gh-ema, line 290: Does this code handle all of the different `SaveMode`s properly? Consider also `Ignore` and `ErrorIfExists`.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: To play it safe, this function creates a new connection and the connection is closed after table existence check. So it is not necessary to change back.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: Excellent catch! This change only affects `OverWrite` and `Append`. `Ignore` and `ErrorIfExists` are handled in `DefaultSource`.  I will add tests and update patch for them.

sfc-gh-ema (approved):

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
```diff
@@ -109,14 +109,17 @@ class DefaultSource(jdbcWrapper: JDBCWrapper)
...
 
-    def tableExists: Boolean = {
-      val conn = jdbcWrapper.getConnector(params)
-      try {
-        jdbcWrapper.tableExists(conn, table.toString)
-      } finally {
-        conn.close()
+    def tableExists: Boolean =
+      if (params.checkTableExistenceInCurrentSchemaOnly) {
+        DefaultJDBCWrapper.tableExistsInCurrentSchema(params, table.toString)
```
sfc-gh-sshankar, line 114: Why DefaultJDBCWrapper ? This is unexpcted . Why not just jdbcWrapper. tableExistsInCurrentSchema

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -609,6 +618,9 @@ object Parameters {
     def quoteJsonFieldName: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_QUOTE_JSON_FIELD_NAME, "true"))
     }
+    def checkTableExistenceInCurrentSchemaOnly: Boolean = {
```
sfc-gh-sshankar, line 621: If this param is true, shouldn't we just always call ```alter session set search_path='$current'```

sfc-gh-mrui: It's doable. But I think it's a better choice to use a new connection.
The problem to use existing connection is that SC needs to get the current value first, and then restore it in the end. SC needs to parse the query result when retrieving the current values. If there is potential interface change, it will break SC. The criteria is lowest risk. I think we should use a new connection.

src/main/scala/net/snowflake/spark/snowflake/DefaultSource.scala
sfc-gh-mrui: In implementation, `jdbcWrapper` is `DefaultJDBCWrapper`. I will update to use `jdbcWrapper `

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -287,10 +287,30 @@ private[snowflake] class JDBCWrapper {
...
     * Returns true if the table already exists in the JDBC database.
+    * By default, if table is an unqualified table name (without schema or database name),
+    * snowflake checks table existence through a search path. For example,
+    * it checks current schema and PUBLIC schema. If the caller intends to check
+    * the table existence in current schema only. The caller needs to alter session to set
+    * search_path as $current explicitly, for example,
+    * conn.createStatement().execute("alter session set search_path='$current'")
+    * or use alternative function: tableExistsInCurrentSchema().
     */
   def tableExists(conn: Connection, table: String): Boolean =
```
sfc-gh-sshankar, line 298: Delete this function, keep only the one below.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -287,10 +287,30 @@ private[snowflake] class JDBCWrapper {
...
     conn.tableExists(table)
 
+  /**
+    * Returns true if the table already exists in the current schema.
+    */
+  private[snowflake] def tableExistsInCurrentSchema(params: MergedParameters, table: String): Boolean = {
```
sfc-gh-sshankar, line 304: Rename to tableExists and check the param here.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -281,16 +281,19 @@ private[io] object StageWriter {
     val writeTableState = new WriteTableState(conn)
 
     try {
+      val tableExists = if (params.checkTableExistenceInCurrentSchemaOnly){
```
sfc-gh-sshankar, line 284: Change all of these assignments to just call DefaultJDBCWrapper.tableExists, where tableExists checks the parameter and uses a new connection

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: Follow it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: I prefer to to keep it because it is used a lot in test code. It needs many test code change. it looks not necessary.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: follow it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-sshankar: Can you create a follow up PR to delete it. It's confusing and error prone to have one version of the function that checks a parameter and another that doesn't.

sfc-gh-mrui: It makes sense. I will clean it in a following PR.

PR #329, opened 2021-02-02
SNOW-262080 Remove redundant function JDBCDriver.tableExists(conn,table)

sfc-gh-mrui:
This is a follow-up PR for https://github.com/snowflakedb/spark-snowflake/pull/328. PR 328 introduced below new function. So `def tableExists(conn: Connection, table: String)` is not necessary any more. But the old function is used in test code. To avoid the PR to have too much change, we introduce this follow-up PR to remove the old function.
The function is not `private[snowflake]`, but the class `JDBCWrapper` is `private[snowflake]`, so it is safe to remove this function.

```
  private[snowflake] def tableExists(params: MergedParameters, table: String): Boolean = {
    val conn = getConnector(params)
    try {
      if (params.checkTableExistenceInCurrentSchemaOnly){
        conn.createStatement().execute("alter session set search_path='$current'")
      }
      conn.tableExists(table)
    } finally {
      conn.close()
    }
  }
```


sfc-gh-bli (approved):

sfc-gh-ema (approved):
LGTM

PR #335, opened 2021-03-01
SNOW-293194 Quote column names when writing with JSON format

sfc-gh-mrui:
Issue description
============
When writing a dataframe with Json format, spark connector specified the column names in the COPY command. For example, 
```
copy into test_table_write_749083442521451968_staging_1100817666 (
  CREATE,TABLE,VIEW,INSERT,TEST )
from ( SELECT_FROM_STAGE ) 
FILE_FORMAT = (TYPE = JSON )
```
But, in this customer case, they are using snowflake key word as column name like CREATE. syntax error happens for COPY command.

Fix
======
The fix is straight forward.  We need to quote the column name in the COPY commnand. For example,
```
copy into test_table_write_749083442521451968_staging_617019349 (
 "CREATE","TABLE","VIEW","INSERT","TEST" )
from ( SELECT_FROM_STAGE ) 
FILE_FORMAT = ( TYPE = JSON)
```

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
@@ -697,7 +697,18 @@ private[io] object StageWriter {
             DefaultJDBCWrapper.resolveTable(conn, table.name, params)
           if (list.isEmpty || list.get.isEmpty) {
             ConstantString("(") + tableSchema.fields
-              .map(_.name)
+              .map(
+                x =>
+                  if (params.quoteJsonFieldName) {
+                    if (params.keepOriginalColumnNameCase) {
+                      Utils.quotedNameIgnoreCase(x.name)
+                    } else {
+                      Utils.ensureQuoted(x.name)
+                    }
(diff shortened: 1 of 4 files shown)


src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -697,7 +697,18 @@ private[io] object StageWriter {
             DefaultJDBCWrapper.resolveTable(conn, table.name, params)
           if (list.isEmpty || list.get.isEmpty) {
             ConstantString("(") + tableSchema.fields
-              .map(_.name)
+              .map(
```
sfc-gh-mrui, line 700: To reviewer:
1. Apply the same quotation strategy like ELSE clause.
2. Reuse existing internal option `params.quoteJsonFieldName` to disable the column name quote if necessary.

sfc-gh-bli (approved):
LGTM

sfc-gh-ema (approved):

sfc-gh-sshankar (approved):
Approval

PR #336, opened 2021-03-02
Support window function avg()/count()/max()/min()/sum() with window frame

sfc-gh-mrui:
In the team meeting, we decided to detect whether the WindowExpression has a WindowFrame. If it doesn’t, we let it pushdown, otherwise not. After future investigation, I found it is not doable.

Window frames require that the data in the window be in a known order. Therefore, the ORDER BY inside the OVER clause is required for window frame syntax, even though that ORDER BY clause is optional in general window function syntax.

So we can’t detect whether a WindowFrame is defined. We can only detect whether there is a ORDER BY.

So, I think we can do this: 
1. For the 5 common aggregate/window functions, add Window Frame for them.
2. Add test case to test them.

This change is low risk because this change only actual enable Window Frame for these five functions. All the others are not affected.


sfc-gh-mrui:
Close it because we decided not to pushdown avg() with window frame. For details refer to https://snowflakecomputing.atlassian.net/browse/SNOW-290292.

PR #338, opened 2021-03-09
SNOW-296738 Disable some Window function pushdown to avoid wrong result.

sfc-gh-mrui:
Issue
====
By design, spark connector intends to support pushdown window function: RowNumber, Rank and DenseRank. 
But, if the function is both aggregate and window function, it is also pushdown.
For that case, Spark connector ignored the Window Frame when pushdown these functions.
This may cause wrong result.

This fix is to disable the pushdown if below 2 conditions are satisfied to avoid potential wrong result:
```
          // Disable window function pushdown if
          // 1. The function are both window function and aggregate function
          // 2. User specifies Window Frame. But there is no way to detect
          //    whether the window function has Window Frame or not.
          //    So we check whether ORDER BY is specified instead.
          //    This may disable the window function which has ORDER BY but
          //    without Window Frame. This is not an issue because it still works.
```

Implementation
===========
1. Add a function: `AggregationStatement.supportWindowFrame()` to determine whether a function is both window and aggregate function, and it supports Window Frame. This information is from https://docs.snowflake.com/en/sql-reference/functions-analytic.html.
2. In `WindowStatement.unapply()`, if the 3 conditions are satisfied, doesn't pushdown it.
3. Two spark migrated test cases: `SFDataFrameWindowFramesSuite` and `SFDataFrameWindowFunctionsSuite` don't actually reading from snowflake. I change to make them run with snowflake.

### DIFF
--- a/src/it/scala/org/apache/spark/sql/SFDataFrameWindowFramesSuite.scala
+++ b/src/it/scala/org/apache/spark/sql/SFDataFrameWindowFramesSuite.scala
@@ -1,14 +1,358 @@
 package org.apache.spark.sql
 
+import org.apache.spark.sql.expressions.Window
+import org.apache.spark.sql.functions._
 import org.apache.spark.sql.snowflake.{SFQueryTest, SFTestData, SFTestSessionBase}
 
 class SFDataFrameWindowFramesSuite
-    extends DataFrameWindowFramesSuite
+    extends SFQueryTest
     with SFTestSessionBase
-    with SFQueryTest
     with SFTestData {
+
+  import SFTestImplicits._
 
   override def spark: SparkSession = getSnowflakeSession()
 
   override protected def blackList: Seq[String] = Seq.empty
+
+  test("lead/lag with empty data frame") {
+    val df = Seq.empty[(Int, String)].toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value")
+
+    checkAnswer(
+      df.select(
+        lead("value", 1).over(window),
+        lag("value", 1).over(window)),
+      Nil)
+  }
+
+  test("lead/lag with positive offset") {
+    val df = Seq((1, "1"), (2, "2"), (1, "3"), (2, "4")).toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value")
+
+    checkAnswer(
+      df.select(
+        $"key",
+        lead("value", 1).over(window),
+        lag("value", 1).over(window)),
+      Row(1, "3", null) :: Row(1, null, "1") :: Row(2, "4", null) :: Row(2, null, "2") :: Nil)
+  }
+
+  test("reverse lead/lag with positive offset") {
+    val df = Seq((1, "1"), (2, "2"), (1, "3"), (2, "4")).toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value".desc)
+
+    checkAnswer(
+      df.select(
+        $"key",
+        lead("value", 1).over(window),
+        lag("value", 1).over(window)),
+      Row(1, "1", null) :: Row(1, null, "3") :: Row(2, "2", null) :: Row(2, null, "4") :: Nil)
+  }
+
+  test("lead/lag with negative offset") {
+    val df = Seq((1, "1"), (2, "2"), (1, "3"), (2, "4")).toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value")
+
+    checkAnswer(
+      df.select(
+        $"key",
+        lead("value", -1).over(window),
+        lag("value", -1).over(window)),
+      Row(1, null, "3") :: Row(1, "1", null) :: Row(2, null, "4") :: Row(2, "2", null) :: Nil)
+  }
+
+  test("reverse lead/lag with negative offset") {
+    val df = Seq((1, "1"), (2, "2"), (1, "3"), (2, "4")).toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value".desc)
+
+    checkAnswer(
+      df.select(
+        $"key",
+        lead("value", -1).over(window),
+        lag("value", -1).over(window)),
+      Row(1, null, "1") :: Row(1, "3", null) :: Row(2, null, "2") :: Row(2, "4", null) :: Nil)
+  }
+
+  test("lead/lag with default value") {
+    val default = "n/a"
+    val df = Seq((1, "1"), (2, "2"), (1, "3"), (2, "4"), (2, "5")).toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value")
+
+    checkAnswer(
+      df.select(
+        $"key",
+        lead("value", 2, default).over(window),
+        lag("value", 2, default).over(window),
+        lead("value", -2, default).over(window),
+        lag("value", -2, default).over(window)),
+      Row(1, default, default, default, default) :: Row(1, default, default, default, default) ::
+        Row(2, "5", default, default, "5") :: Row(2, default, "2", "2", default) ::
+        Row(2, default, default, default, default) :: Nil)
+  }
+
+  test("rows/range between with empty data frame") {
+    val df = Seq.empty[(String, Int)].toDF("key", "value")
+    val window = Window.partitionBy($"key").orderBy($"value")
+
+    checkAnswer(
+      df.select(
+        $"key",
+        first("value").over(
+          window.rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)),
+        first("value").over(
+          window.rangeBetween(Window.unboundedPreceding, Window.unboundedFollowing))),
+      Nil)
+  }
+
+  test("rows between should accept int/long values as boundary") {
+    val df = Seq((1L, "1"), (1L, "1"), (2147483650L, "1"), (3L, "2"), (2L, "1"), (2147483650L, "2"))
+      .toDF("key", "value")
+
+    checkAnswer(
+      df.select(
+        $"key",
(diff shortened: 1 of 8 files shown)


src/it/scala/org/apache/spark/sql/SFDataFrameWindowFramesSuite.scala
```diff
@@ -1,14 +1,358 @@
...
     with SFTestSessionBase
-    with SFQueryTest
     with SFTestData {
 
+  import SFTestImplicits._
+
   override def spark: SparkSession = getSnowflakeSession()
 
   override protected def blackList: Seq[String] = Seq.empty
+
```
sfc-gh-mrui, line 17: To reviewers:
`SFDataFrameWindowFramesSuite` and `SFDataFrameWindowFunctionsSuite` don't actually run with snowflake database. I copy all `test` from `DataFrameWindowFramesSuite` and `DataFrameWindowFunctionsSuite`, and then make them run with snowflake.

src/it/scala/org/apache/spark/sql/snowflake/SFQueryTest.scala
```diff
@@ -62,6 +62,19 @@ object SFQueryTest {
     case (a: java.math.BigDecimal, b: Float) =>
       (a.toString.toFloat - b).abs < 0.0001
     case (a: java.math.BigDecimal, b: java.math.BigDecimal) => a.equals(b)
```
sfc-gh-mrui, line 64: to reviewers:
`scala.math.BigDecimal` and `java.math.BigDecimal` are two different classes.
Some testing data is using `scala.math.BigDecimal`, this function needs to be updated to handle it like `java.math.BigDecimal`.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -1729,7 +1729,8 @@ case class InternalGcsStorage(param: MergedParameters,
     val metadatas = generateFileTransferMetadatas(
       data, format, compress, directory, data.getNumPartitions)
```
sfc-gh-mrui, line 1731: To Reviewer:
After enabling these tests, I find one issue for writing an empty DataFrame (partition count is 0) to snowflake on GCP.
Separately Jira is filed. The fix is one line fix, so, I just bundle the fix here because current test can test it.
  https://snowflakecomputing.atlassian.net/browse/SNOW-296768

sfc-gh-bli: to write a empty data frame, we can create a dummy row first to make it not empty, and then filter this dummy row out, to make it empty again.

sfc-gh-bli (approved):
LGTM

sfc-gh-ema:
What happens if the function has a window frame, is a window function, but is NOT an aggregate function? What is the behavior of the connector in that case?

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
sfc-gh-mrui: It looks there are two cases for empty data frame.
Case 1: The partition count for the data frame is > 0, but the row counts for each partition is 0.
Case 2: The partition count is 0.

We have test cases for case 1. But we don't have test case for case 2. Below code can generate data frame without any partition. 
```
    val df = Seq.empty[(Int, String)].toDF("key", "value")

```

sfc-gh-mrui:
> What happens if the function has a window frame, is a window function, but is NOT an aggregate function? What is the behavior of the connector in that case?

@sfc-gh-ema I think this will not happen.
In `WindowStatement`, by default, the function is handled like below. It calls `convertStatement(func, fields)` for non-listed functions. But, if the function is an aggregate function, `convertStatement(func, fields)` will call `AggregationStatement` to parse the aggregate function. Finally, the function is pushdown. This case doesn't happen for other functions.
```
          case _ =>
            convertStatement(func, fields) + " OVER " + windowBlock(
              spec,
              fields,
              useWindowFrame = false
            )
```

In addition, according to https://docs.snowflake.com/en/sql-reference/functions-analytic.html, except aggregate functions,  only some Rank-related functions supports Window Frame. They are not pushdown anyway.

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
```diff
@@ -1,22 +1,810 @@
...
+
+    withSQLConf(SQLConf.WINDOW_EXEC_BUFFER_IN_MEMORY_THRESHOLD.key -> "1",
+      SQLConf.WINDOW_EXEC_BUFFER_SPILL_THRESHOLD.key -> "1") {
+      assertSpilled(sparkContext, "select") {
+        df.select($"key", sum("value").over(window)).collect()
+      }
+    }
+  }
+
+//  test("SPARK-21258: complex object in combination with spilling") {
```
sfc-gh-sshankar, line 640: What's all this stuff commented out ?

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
```diff
@@ -1,22 +1,810 @@
...
+    checkAnalysisError(
+      sql(
+        s"""SELECT a, MAX(b)
+           |FROM testData2
+           |GROUP BY a
+           |HAVING SUM(b) = 5 AND RANK() OVER(ORDER BY a) = 1""".stripMargin),
+      "HAVING")
+  }
+
+//  test("window functions in multiple selects") {
```
sfc-gh-sshankar, line 730: Again why is stuff commented out ?

src/it/scala/org/apache/spark/sql/snowflake/SFTestWrapperSparkSession.scala
```diff
@@ -28,6 +28,10 @@ class SFTestWrapperSparkSession(sc: SparkContext, sfConfigs: Map[String, String]
       .options(sfConfigs)
       .option("dbtable", tableName)
       .save()
+    if (df.count() == 0) {
```
sfc-gh-sshankar, line 31: Why do you need this ?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/WindowStatement.scala
```diff
@@ -46,6 +47,14 @@ private[querygeneration] object WindowStatement {
               useWindowFrame = true
             )
 
+          // Disable window function pushdown if
+          // 1. The function are both window function and aggregate function
+          // 2. The function support Window Frame
+          // 3. The user specify ORDER-BY clause
+          case agg: AggregateExpression
+            if AggregationStatement.supportWindowFrame(agg) && spec.orderSpec.nonEmpty =>
```
sfc-gh-sshankar, line 55: Why can;t you check if there is in fact a window function ? Isn;t that easier ?

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
sfc-gh-mrui: This case is commented out, because is use `input.toDS()`, We don't support it yet.

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
sfc-gh-mrui: This is commented out because `stripAQEPlan(select.queryExecution.executedPlan)` is used, we don't support it yet. And it looks it is spark internal, it is not necessary for us.

src/it/scala/org/apache/spark/sql/SFDataFrameWindowFunctionsSuite.scala
sfc-gh-sshankar: block comments please, and add a comment as to why ?

src/it/scala/org/apache/spark/sql/snowflake/SFTestWrapperSparkSession.scala
sfc-gh-sshankar: That's weird. Also what you have here is extremely inefficient. You are evaluating the dataframe twice. Instead just do a create if not exists on the table.
Is th efact that a table is not created a bug in our connector ?

src/it/scala/org/apache/spark/sql/snowflake/SFQueryTest.scala
sfc-gh-ema: Why are all of these additions needed?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/WindowStatement.scala
sfc-gh-ema: I still don't understand why we care about `supportWindowFrame`. If there is a window frame in Spark, even though Snowflake may not support it, why do we want to default to the `case _` behavior? Shouldn't we just bail out with `null` in that case?

src/it/scala/org/apache/spark/sql/snowflake/SFTestWrapperSparkSession.scala
sfc-gh-ema: `df.write` in `Overwrite` mode should still create the table with zero rows, I think. If it doesn't, that is a bug.

sfc-gh-ema (approved):

PR #340, opened 2021-03-18
SNOW-358401 Support spark 3.1

sfc-gh-mrui:
Spark 3.1 has 3 change affects SC pushdown. Below are what's the change and how SC process it.

1. Change constructor for `Union`. It has 2 new parameters: `byName` and `allowMissingCol`. 
    Snowflake doesn’t support this spark featured UNION. SC only pushdown UNION if these 2 parameters are FALSE.
```
case class Union(
    children: Seq[LogicalPlan],
    byName: Boolean = false,
    allowMissingCol: Boolean = false)
```

2. Change constructor for UnaryMinus. It has a new parameters `failOnError`.
    It defines the behaviour for `Minus` operator for spark when error happens. SC can ignore it.
```
case class UnaryMinus(
    child: Expression,
    failOnError: Boolean = SQLConf.get.ansiEnabled)
```

3. Change constructor for Rand . It has a new parameter hideSeed.
    `hideSeed` is used to avoid printing `seed` in the plan. SC can ignore it.
```
case class Rand(child: Expression, hideSeed: Boolean = false)

override def sql: String = {
  s"rand(${if (hideSeed) "" else child.sql})"
}
```

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
@@ -123,7 +123,7 @@ object SnowflakeFailMessage {
   final val FAIL_PUSHDOWN_SET_TO_EXPR = "pushdown failed in setToExpr"
   final val FAIL_PUSHDOWN_AGGREGATE_EXPRESSION = "pushdown failed for aggregate expression"
   final val FAIL_PUSHDOWN_UNSUPPORTED_CONVERSION = "pushdown failed for unsupported conversion"
-  final val FAIL_PUSHDOWN_UNSUPPORTED_UNION = "pushdown failed for Spark featured UNION"
+  final val FAIL_PUSHDOWN_UNSUPPORTED_UNION = "pushdown failed for Spark feature: UNION by name"
 }
 
 class SnowflakePushdownUnsupportedException(message: String,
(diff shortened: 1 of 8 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
```diff
@@ -123,6 +123,7 @@ object SnowflakeFailMessage {
   final val FAIL_PUSHDOWN_SET_TO_EXPR = "pushdown failed in setToExpr"
   final val FAIL_PUSHDOWN_AGGREGATE_EXPRESSION = "pushdown failed for aggregate expression"
   final val FAIL_PUSHDOWN_UNSUPPORTED_CONVERSION = "pushdown failed for unsupported conversion"
+  final val FAIL_PUSHDOWN_UNSUPPORTED_UNION = "pushdown failed for Spark featured UNION"
```
sfc-gh-ema, line 126: nit: "Spark feature: UNION"

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/QueryBuilder.scala
```diff
@@ -193,8 +193,21 @@ private[querygeneration] class QueryBuilder(plan: LogicalPlan) {
           }
         }
 
-      case Union(children) =>
-        Some(UnionQuery(children, alias.next))
+      // From Spark 3.1, Union has 3 parameters
```
sfc-gh-ema, line 196: Could you write an explanation (as a comment) as to why this isn't supported in Snowflake? What do `byName` and `allowMissingCol` do?

sfc-gh-ema (approved):

sfc-gh-bli (approved):
LGTM

PR #341, opened 2021-03-22
SNOW-297134 Support to write a DataFrame without any partition.

sfc-gh-mrui:
Description
=========
Below DataFrame doesn't have any partition. Currently, the WRITE operation is skipped when writing it to snowflake.
`val emptyDf = Seq.empty[(Int, String)].toDF("key", "value")
`

Analysis
======
I have confirmed to write this DataFrame as CSV file and Parquet file. It generates an empty file for WRITE operation.
So, we will not skip the WRITE operation either. For example,
1. If the target doesn't exists, it will create an empty table.
2. If the target table exists and write mode is OverWrite, a new table is created to overwrite the old table.

Implementation
===========
1. If the dataframe has no partitions
   a) If save mode is Append, executes `CREATE  TABLE IF  NOT EXISTS ...`  to create table.
   b) If write mode is OverWrite, executes `CREATE OR REPLACE TABLE ...`  to create table.
2. Introduce below internal option to disable this fix to play it safe.
```
  // Internal option to skip write operation when writing a DataFrame
  // without any partitions. By default, it it is 'false'.
  // This option may be removed without any notice in any time.
  // Introduced since Spark Connector 2.8.5
  val PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME: String = knownParam(
    "internal_skip_write_when_writing_empty_dataframe"
  )
```

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -162,6 +162,13 @@ object Parameters {
   val PARAM_INTERNAL_CHECK_TABLE_EXISTENCE_IN_CURRENT_SCHEMA_ONLY: String = knownParam(
     "internal_check_table_existence_in_current_schema_only"
   )
+  // Internal option to skip write operation when writing a DataFrame
+  // without any partitions. By default, it it is 'false'.
+  // This option may be removed without any notice in any time.
+  // Introduced since Spark Connector 2.8.5
+  val PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME: String = knownParam(
+    "internal_skip_write_when_writing_empty_dataframe"
+  )
 
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
   val MIN_S3_MAX_FILE_SIZE = 1000000
@@ -620,6 +627,9 @@ object Parameters {
     }
     def checkTableExistenceInCurrentSchemaOnly: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_CHECK_TABLE_EXISTENCE_IN_CURRENT_SCHEMA_ONLY, "true"))
+    }
+    def skipWriteWhenWritingEmptyDataFrame: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME, "false"))
     }
 
     /**
--- a/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
@@ -225,12 +225,23 @@ private[io] object StageWriter {
           format,
           fileUploadResults
         )
-      } else {
+      } else if (params.skipWriteWhenWritingEmptyDataFrame) {
         log.info(
           s"""${SnowflakeResultSetRDD.MASTER_LOG_PREFIX}:
              | Skip to execute COPY INTO TABLE command because
              | no file is uploaded.
              |""".stripMargin.filter(_ >= ' '))
+      } else {
+        writeToTable(
+          conn,
+          schema,
+          saveMode,
+          params,
+          "dummy_prefix_for_write_empty_partition",
+          stage,
+          format,
+          fileUploadResults
+        )
       }
       val endTime = System.currentTimeMillis()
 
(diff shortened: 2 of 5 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -162,6 +162,13 @@ object Parameters {
   val PARAM_INTERNAL_CHECK_TABLE_EXISTENCE_IN_CURRENT_SCHEMA_ONLY: String = knownParam(
     "internal_check_table_existence_in_current_schema_only"
   )
+  // Internal option to skip write operation when writing a DataFrame
+  // without any partitions. By default, it it is 'false'.
+  // This option may be removed without any notice in any time.
+  // Introduced since Spark Connector 2.8.5
+  val PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME: String = knownParam(
+    "internal_skip_write_when_writing_empty_dataframe"
+  )
```
sfc-gh-bli, line 171: why we need this parameter? this PR fixes a bug right? it is not a new feature

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
```diff
@@ -225,12 +225,23 @@ private[io] object StageWriter {
           format,
           fileUploadResults
         )
-      } else {
+      } else if (params.skipWriteWhenWritingEmptyDataFrame) {
```
sfc-gh-bli, line 228: I don't understand why this design is necessary. My idea is, if a DataFrame has empty partition, then we can directly create a table without invoke COPY.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: I agree this is bug fix. Even if it is bug fix, it is a behaviour change.
Our current strategy is to introduce an internal option to disable the behavior change.
If any customer depends on the OLD behaviour, they can use this internal option.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: Yes. We can do that.
If the write mode is append, we will `CREATE TABLE if not exist ...`
If the write mode is overwrite, we will execute `CREATE OR REPLAE  table ...`

I will do that.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-bli: is it a behavior change? 
Change from not support empty DataFrame to support empty DataFrame is not a behavior change.
It is a bug fix or kinds of new feature.

sfc-gh-bli: @sfc-gh-sshankar @sfc-gh-ema what do you think?

sfc-gh-ema: I think if we use parameters like this, we need to have a long-term plan to remove these, or else we will end up with too many parameters.

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-ema: We log only if this parameter is `true`? Don't we still need to create the table?

sfc-gh-mrui: The intention for the `if (params.skipWriteWhenWritingEmptyDataFrame)` is to go to the OLD behaviour if necessary. So we don't create table for it.
By default, it is false, so it will go to `createTable()` in `else` clause

sfc-gh-bli (approved):
File A Jira to remove this parameter in future

src/main/scala/net/snowflake/spark/snowflake/io/StageWriter.scala
sfc-gh-mrui: let's keep this parameter to play it safe. In the same time, Jira https://snowflakecomputing.atlassian.net/browse/SNOW-330109 is filed to clean-up the internal option in the future.

sfc-gh-mrui:
> File A Jira to remove this parameter in future
https://snowflakecomputing.atlassian.net/browse/SNOW-330109 is filed for this.

sfc-gh-ema (approved):

PR #344, opened 2021-03-31
SNOW-330736 Enable to pushdown EqualNullSafe: operator '<=>'

sfc-gh-mrui:
The issue is because Spark `EqualNullSafe:  <=> `is pushdown to snowflake as `<=>`. But snowflake doesn’t support this operator.
Snowflake has equivalent function `EQUAL_NULL(left, right)`. So SC can pushdown `EqualNullSafe` as `EQUAL_NULL(left, right)`.


sfc-gh-bli (approved):
LGTM

sfc-gh-ema (approved):

PR #346, opened 2021-04-01
SNOW-332466 Disable pushdown for converting from DATE/TIMESTAMP to NUMBER

sfc-gh-mrui:
This change doesn't affect functionality.
If the conversion is pushdown to snowflake, snowflake will raise conversion error like below. SC will retry without pushdown. The reading will succeed in the end.
But this change will avoid users see a lot of failed query.
```
SQL compilation error:
invalid type [CAST(SUBQUERY_0.D AS FLOAT)] for parameter 'TO_DOUBLE'
```

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
@@ -46,6 +46,7 @@ private[querygeneration] object MiscStatement {
       case Cast(child, t, _) =>
         getCastType(t) match {
           case Some(cast) =>
+            raiseExceptionIfCastIsNotSupported(child.dataType, t)
             ConstantString("CAST") +
               blockStatement(convertStatement(child, fields) + "AS" + cast)
           case _ => convertStatement(child, fields)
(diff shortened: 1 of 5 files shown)


sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
```diff
@@ -159,4 +160,33 @@ private[querygeneration] object MiscStatement {
...
+      // QueryBuilder.treeRoot and a telemetry message will be sent if
+      // there are any snowflake tables in the query.
+      throw new SnowflakePushdownUnsupportedException(
+        SnowflakeFailMessage.FAIL_PUSHDOWN_UNSUPPORTED_CONVERSION,
+        s"Don't support to convert $srcType column to $targetType type",
+        "",
+        false
+      )
+    }
+  }
```
sfc-gh-bli, line 190: Can be simplified to two lines
```scala
(srcType, targetType) match {
case (DateType, NumericType) => throw Exception
case (TimestampType, NumericType) => throw Exception
}
```

sfc-gh-bli: And I don't think you need this separate function, move it inline

sfc-gh-mrui: I just tried, but it looks `NumericType` is spark private. SC can't access it.
`private[spark] object NumericType extends AbstractDataType {...}`

sfc-gh-mrui: If it can't be simplified as you mentioned, I think it is better to use a separate function.

sfc-gh-bli (approved):

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
sfc-gh-mrui: it is simplified as below:
```
      case Cast(child, t, _) =>
        getCastType(t) match {
          case Some(cast) =>
            /** For known unsupported data conversion, raise exception to break the pushdown process.
              * For example, snowflake doesn't support to convert DATE/TIMESTAMP to NUMBER
              */
            (child.dataType, t) match {
              case (_: DateType | _: TimestampType,
              _: IntegerType | _: LongType | _: FloatType | _: DoubleType | _: DecimalType) => {
                // This exception will not break the connector. It will be caught in
                // QueryBuilder.treeRoot and a telemetry message will be sent if
                // there are any snowflake tables in the query.
                throw new SnowflakePushdownUnsupportedException(
                  SnowflakeFailMessage.FAIL_PUSHDOWN_UNSUPPORTED_CONVERSION,
                  s"Don't support to convert ${child.dataType} column to $t type",
                  "",
                  false)
              }
              case _ =>
            }

            ConstantString("CAST") +
              blockStatement(convertStatement(child, fields) + "AS" + cast)
          case _ => convertStatement(child, fields)
        }
```

PR #357, opened 2021-05-19
SNOW-346101 Set client info with JDBC properties instead of system property

sfc-gh-mrui:
Currently, Spark connector uses system property to set CLIENT INFO. But this is not reliable in some cases, for example, the user may have no privilege to set system property.

From JDBC 3.13.3, JDBC supports to set CLIENT INFO with JDBC properties.
So, update Spark connector to set it in JDBC properties.

In the same time, bump JDBC version to latest 3.13.3.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -24,6 +24,7 @@ import java.util.Properties
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.{Executors, ThreadFactory}
 
+import net.snowflake.client.core.SFSessionProperty
 import net.snowflake.client.jdbc.telemetry.{Telemetry, TelemetryClient}
 import net.snowflake.spark.snowflake.DefaultJDBCWrapper.DataBaseOperations
 import net.snowflake.spark.snowflake.Parameters.MergedParameters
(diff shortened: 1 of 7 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -199,9 +200,11 @@ private[snowflake] class JDBCWrapper {
            |""".stripMargin.filter(_ >= ' '))
     }
 
-    // Important: Set "snowflake.client.info" is very important!
+    // Important: Set client_info is very important!
     // For more details, refer to PROPERTY_NAME_OF_CONNECTOR_VERSION
-    System.setProperty("snowflake.client.info", Utils.getClientInfoString())
+    // NOTE: From JDBC 3.13.3, the client info can be set with JDBC properties
+    // instead of system property: "snowflake.client.info".
+    jdbcProperties.put(SFSessionProperty.CLIENT_INFO.getPropertyKey, Utils.getClientInfoString())
```
sfc-gh-mrui, line 207: To reviewers:
`SFSessionProperty.CLIENT_INFO` is a new property from JDBC 3.13.3, which is used to set CLIENT INFO with JDBC properties.

sfc-gh-bli: So, that is a JDBC change, we just apply this change?

sfc-gh-bli (approved):

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
sfc-gh-mrui: Now, JDBC will check this CLIENT_INFO first. If it is not, it checks System property so that some old spark connector can use new JDBC.

The story is:
1. We found this problem and discuss with JDBC team. https://snowflake.slack.com/archives/GCA15K0QH/p1618940626292300
2. They change JDBC API to support set CLIENT INFO with JDBC properties.
3. Now, we change SC to use the new API.

PR #366, opened 2021-07-19
SNOW-370053 Enable SORT + LIMIT test case for Large data on AWS sfctest0

sfc-gh-mrui:
The new test case added for SORT + LIMIT fails in the regression test.
The reason is that the test needs there is a large table on the test account.
but there is no such a large table on preprod3 test account.
So, the new test needs to be disabled for regression test.

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
@@ -161,11 +161,12 @@ class SimpleNewPushdownIntegrationSuite extends IntegrationSuiteBase {
   }
 
   test("LIMIT and SORT with large table") {
-    val accountName = System.getenv(SNOWFLAKE_TEST_ACCOUNT)
-    val isAWS = accountName == null || accountName.equals("aws")
-    // This test case can only be enabled on AWS test account
+    val deploymentName = System.getenv(SNOWFLAKE_TEST_ACCOUNT)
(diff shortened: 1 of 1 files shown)


src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
```diff
@@ -161,11 +161,12 @@ class SimpleNewPushdownIntegrationSuite extends IntegrationSuiteBase {
...
-    val accountName = System.getenv(SNOWFLAKE_TEST_ACCOUNT)
-    val isAWS = accountName == null || accountName.equals("aws")
-    // This test case can only be enabled on AWS test account
+    val deploymentName = System.getenv(SNOWFLAKE_TEST_ACCOUNT)
+    val isAWS = deploymentName == null || deploymentName.equals("aws")
+    // This test case can only be enabled on AWS test account sfctest0
     // because the large test table only is accessible on it.
-    if (isAWS) {
+    val accountName = connectorOptionsNoTable("sfurl").toLowerCase
+    if (isAWS && accountName.contains("sfctest0")) {
```
sfc-gh-ema, line 169: Instead of doing this, can we check for the existence of the required table?

sfc-gh-mrui: Good idea. Update test case.

sfc-gh-ema (approved):

sfc-gh-bli (approved):
I remember I imported that shared DB on preprod3. anyway, if you want that shared db, you can create it by yourself.

PR #368, opened 2021-07-21
SNOW-394972 Get schema by preparing the statement instead of executing it

sfc-gh-mrui:
SNOW-394972

Description
=========
Currently, when spark connector needs to get the metadata for table or query, it executes `"SELECT * from ($TABLE_OR_QUERY) where 1 = 2"`. Change to prepare the statement only for efficiency.

Implementation
============
1. Create a new function  `SnowflakeSQLStatement.prepareStatement()` by refactoring `SnowflakeSQLStatement.execute()`.
2. Enhance `DefaultJDBCWrapper.tableMetaDataFromStatement()` to call `prepareStatement()` to get the ResultSetMetadata.

NOTE:
1. There are multiple places to resolve table or query schema. I have confirmed all of them are finally call `tableMetaDataFromStatement().
2. No new test cases are needed since it has been covered by existed test cases.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -556,7 +556,7 @@ private[snowflake] object DefaultJDBCWrapper extends JDBCWrapper {
       bindVariableEnabled: Boolean = true
     ): ResultSetMetaData =
       (ConstantString("select * from") + statement + "where 1 = 0")
-        .execute(bindVariableEnabled)(connection)
+        .prepareStatement(bindVariableEnabled)(connection)
         .getMetaData
 
     def tableSchema(name: String, params: MergedParameters): StructType =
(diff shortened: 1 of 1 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -556,7 +556,7 @@ private[snowflake] object DefaultJDBCWrapper extends JDBCWrapper {
       bindVariableEnabled: Boolean = true
     ): ResultSetMetaData =
       (ConstantString("select * from") + statement + "where 1 = 0")
-        .execute(bindVariableEnabled)(connection)
+        .prepareStatement(bindVariableEnabled)(connection)
```
sfc-gh-ema, line 559: If we're just preparing the statement without executing it, we don't need the...`select ... where 1=0` wrapper anymore, right? We can just run `prepare` on the statement itself?

sfc-gh-mrui: There are some use cases that the pass-in parameter is table name only.
If we don't wrap it here, all the caller needs to wrap it.
So I think current fix is easier and low risk.

sfc-gh-ema (approved):

sfc-gh-bli (approved):

PR #378, opened 2021-10-06
SNOW-474215 Add 3 options to set 3 timestamp output formats

sfc-gh-mrui:
SNOW-474215

Issue description
============
CTC raises a concern that SC always executes an `alter session` to set timestamp formats when a JDBC connection is created even if they think they don't need these settings. For example,
```
alter session set timezone = 'GMT' ,
 timestamp_ntz_output_format = 'YYYY-MM-DD HH24:MI:SS.FF3',
 timestamp_ltz_output_format = 'TZHTZM YYYY-MM-DD HH24:MI:SS.FF3',
 timestamp_tz_output_format = 'TZHTZM YYYY-MM-DD HH24:MI:SS.FF3' ; 
```

Fix
==
Add 3 spark connector options to set these 3 output format.
1. Their default value is current settings, so it doesn't break any customer
2. User has a choice to set these options if they need.
3. If these 3 options and `sfTimezone` are set as `sf_current`, the `alter session` command can be skipped. This can satisfy CTC's use case.
```
option_name -> default_value
========================
timestamp_ntz_output_format -> "YYYY-MM-DD HH24:MI:SS.FF3"
timestamp_ltz_output_format -> "TZHTZM YYYY-MM-DD HH24:MI:SS.FF3"
timestamp_tz_output_format  -> "TZHTZM YYYY-MM-DD HH24:MI:SS.FF3"
```

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -111,6 +111,10 @@ object Parameters {
     "partition_size_in_mb"
   )
   val PARAM_TIME_OUTPUT_FORMAT: String = knownParam("time_output_format")
+  val PARAM_TIMESTAMP_NTZ_OUTPUT_FORMAT: String = knownParam("timestamp_ntz_output_format")
+  val PARAM_TIMESTAMP_LTZ_OUTPUT_FORMAT: String = knownParam("timestamp_ltz_output_format")
+  val PARAM_TIMESTAMP_TZ_OUTPUT_FORMAT: String = knownParam("timestamp_tz_output_format")
+  val TIMESTAMP_OUTPUT_FORMAT_SF_CURRENT: String = "sf_current"
   val PARAM_JDBC_QUERY_RESULT_FORMAT: String = knownParam(
     "jdbc_query_result_format"
   )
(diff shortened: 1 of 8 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -111,6 +111,10 @@ object Parameters {
     "partition_size_in_mb"
   )
   val PARAM_TIME_OUTPUT_FORMAT: String = knownParam("time_output_format")
+  val PARAM_TIMESTAMP_NTZ_OUTPUT_FORMAT: String = knownParam("timestamp_ntz_output_format")
+  val PARAM_TIMESTAMP_LTZ_OUTPUT_FORMAT: String = knownParam("timestamp_ltz_output_format")
+  val PARAM_TIMESTAMP_TZ_OUTPUT_FORMAT: String = knownParam("timestamp_tz_output_format")
+  val TIMESTAMP_OUTPUT_FORMAT_SF_CURRENT: String = "sf_current"
```
sfc-gh-mrui, line 117: If user set these 3 options as `sf_current`, Spark connector will not set the option for the user.

If users set these 3 options and sfTimezone as `sf_current`, Spark Connector will skip the `alter session` command to set `timestamp and timezone` format.

PR #387, opened 2021-10-28
Support to count large table

### LINKED ISSUE #376: Predicate pushdown of COUNT fails on large table
I'm using `spark-snowflake` connector version `2.9.1` with a `SNAPSHOT` version of Spark compiled using the latest `spark` repository commits on Java `11.0.9`, Scala `2.12.14`, and Snowflake `5.32.4`. It seems that when the `spark-snowflake` connector pushes down a `.count()` operation into `SELECT COUNT(*)` in the actual Snowflake query, an error `java.lang.IllegalArgumentException: More than Int.MaxValue elements.` can trigger. This table is 4.9 billion rows in size, which does exceed `IntMax`, which is about 2.1 billion. Yet since this is a single row we're asking Snowflake to return, the number of rows should be immaterial to how the underlying computation functions in both Snowflake and Spark.

Note that if I re-write the query manually to use `COUNT(*)`, _the correct result is returned_. In addition, if I use `.persist` before `.show` on the `DataFrame`, a subsequent call to `.count()` _also works correctly_. In this second case, Spark is fetching all the rows and persisting them to local disk. The subsequent call to `.count` will use Spark exclusively for calculating the number of rows, rather than any predicate pushdown to Snowflake's engine. These workarounds suggest there could be a bug somewhere in the `spark-snowflake` connector predicate pushdown code, rather than a bug in Spark itself.

One caveat is I'm using the most recent Spark commits & compiling them from scratch, so I'm not using a version of Spark that's been certified to work with `spark-snowflake`. I've observed this error more than once over the last few months & usually work around it with a manual re-write, so I don't think any very recent commits could possibly be responsible. The stack trace also indicates the error occurs inside of a low-level Scala function `scala.collection.immutable.NumericRange`, so I suspect this issue is either present in regular Spark releases or will be present in the upcoming Spark 3.2 release.

**Details**

The code I'm executing in the `spark-shell` REPL is:
```
val exampleQuery = s"""
        SELECT A, B, C
        FROM DB.SCHEMA.MY_TABLE
        WHERE D = true AND B IS NOT NULL AND C IS NOT NULL;
        """

val myDF = spark.sqlContext.read
  .options(sfOptions)
  .option("query", exampleQuery)
  .load()
```

Here, `myDF` is the `DataFrame` we retrieve, `sfOptions` contains authentication data for our Snowflake instance, and `DB.SCHEMA.MY_TABLE` is a table containing 4.9 billion rows.

When using `.show` on the `DataFrame`, the query successfully executes in Snowflake and a few rows are shown as desired. When I check our Snowflake UI, I can see the actual Snowflake query executed is:
```
SELECT "A", "B", "C" FROM (SELECT A, B, C
        FROM DB.SCHEMA.MY_TABLE
        WHERE D = true AND B IS NOT NULL AND C IS NOT NULL) 
```

When I call `myDF.count()`, the operation gets pushed down into the Snowflake query as desired. The actual query executed on Snowflake is now:
```
SELECT count(*) FROM ( SELECT A, B, C
        FROM DB.SCHEMA.MY_TABLE
        WHERE D = true AND B IS NOT NULL AND C IS NOT NULL ) 
```

However, instead of returning 4.9 billion as the count, the command runs for a little while before erroring out with the following stacktrace:
```
java.lang.IllegalArgumentException: More than Int.MaxValue elements.
  at scala.collection.immutable.NumericRange$.check$1(NumericRange.scala:319)
  at scala.collection.immutable.NumericRange$.count(NumericRange.scala:329)
  at scala.collection.immutable.NumericRange.numRangeElements$lzycompute(NumericRange.scala:54)
  at scala.collection.immutable.NumericRange.numRangeElements(NumericRange.scala:53)
  at scala.collection.immutable.NumericRange.length(NumericRange.scala:56)
  at org.apache.spark.rdd.ParallelCollectionRDD$.slice(ParallelCollectionRDD.scala:143)
  at org.apache.spark.rdd.ParallelCollectionRDD.getPartitions(ParallelCollectionRDD.scala:96)
  at org.apache.spark.rdd.RDD.$anonfun$partitions$2(RDD.scala:292)
  at scala.Option.getOrElse(Option.scala:189)
  at org.apache.spark.rdd.RDD.partitions(RDD.scala:288)
  at org.apache.spark.rdd.MapPartitionsRDD.getPartitions(MapPartitionsRDD.scala:49)
  at org.apache.spark.rdd.RDD.$anonfun$partitions$2(RDD.scala:292)
  at scala.Option.getOrElse(Option.scala:189)
  at org.apache.spark.rdd.RDD.partitions(RDD.scala:288)
  at org.apache.spark.rdd.MapPartitionsRDD.getPartitions(MapPartitionsRDD.scala:49)
  at org.apache.spark.rdd.RDD.$anonfun$partitions$2(RDD.scala:292)
  at scala.Option.getOrElse(Option.scala:189)
  at org.apache.spark.rdd.RDD.partitions(RDD.scala:288)
  at org.apache.spark.rdd.MapPartitionsRDD.getPartitions(MapPartitionsRDD.scala:49)
  at org.apache.spark.rdd.RDD.$anonfun$partitions$2(RDD.scala:292)
  at scala.Option.getOrElse(Option.scala:189)
  at org.apache.spark.rdd.RDD.partitions(RDD.scala:288)
  at org.apache.spark.rdd.RDD.getNumPartitions(RDD.scala:308)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeExec.mapOutputStatisticsFuture$lzycompute(ShuffleExchangeExec.scala:140)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeExec.mapOutputStatisticsFuture(ShuffleExchangeExec.scala:139)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeLike.$anonfun$submitShuffleJob$1(ShuffleExchangeExec.scala:68)
  at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:222)
  at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
  at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:219)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeLike.submitShuffleJob(ShuffleExchangeExec.scala:68)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeLike.submitShuffleJob$(ShuffleExchangeExec.scala:67)
  at org.apache.spark.sql.execution.exchange.ShuffleExchangeExec.submitShuffleJob(ShuffleExchangeExec.scala:115)
  at org.apache.spark.sql.execution.adaptive.ShuffleQueryStageExec.shuffleFuture$lzycompute(QueryStageExec.scala:170)
  at org.apache.spark.sql.execution.adaptive.ShuffleQueryStageExec.shuffleFuture(QueryStageExec.scala:170)
  at org.apache.spark.sql.execution.adaptive.ShuffleQueryStageExec.doMaterialize(QueryStageExec.scala:172)
  at org.apache.spark.sql.execution.adaptive.QueryStageExec.materialize(QueryStageExec.scala:82)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.$anonfun$getFinalPhysicalPlan$5(AdaptiveSparkPlanExec.scala:256)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.$anonfun$getFinalPhysicalPlan$5$adapted(AdaptiveSparkPlanExec.scala:254)
  at scala.collection.Iterator.foreach(Iterator.scala:943)
  at scala.collection.Iterator.foreach$(Iterator.scala:943)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
  at scala.collection.IterableLike.foreach(IterableLike.scala:74)
  at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
  at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.$anonfun$getFinalPhysicalPlan$1(AdaptiveSparkPlanExec.scala:254)
  at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.getFinalPhysicalPlan(AdaptiveSparkPlanExec.scala:226)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.withFinalPlanUpdate(AdaptiveSparkPlanExec.scala:365)
  at org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec.executeCollect(AdaptiveSparkPlanExec.scala:338)
  at org.apache.spark.sql.Dataset.$anonfun$count$1(Dataset.scala:3039)
  at org.apache.spark.sql.Dataset.$anonfun$count$1$adapted(Dataset.scala:3038)
  at org.apache.spark.sql.Dataset.$anonfun$withAction$1(Dataset.scala:3733)
  at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$5(SQLExecution.scala:103)
  at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:163)
  at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:90)
  at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:775)
  at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:64)
  at org.apache.spark.sql.Dataset.withAction(Dataset.scala:3731)
  at org.apache.spark.sql.Dataset.count(Dataset.scala:3038)
  ... 47 elided
```

dohongdayi:
fixes #376

SparkContext.parallelize() doesn't support Int64, instead of which we can use SparkContext.range()

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
@@ -1193,6 +1193,17 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
     assert(count == LARGE_TABLE_ROW_COUNT)
   }
 
+  test("count super large table") {
+    if (!skipBigDataTest) {
+      val expectedCount = Int.MaxValue + 1L
+      val actualCount = getRowCount(
+        s"table(generator(rowcount => $expectedCount))",
+        thisConnectorOptionsNoTable
+      )
+      assert(actualCount == expectedCount)
+    }
+  }
+
   // Copy one table from AWS account to GCP account
   ignore("copy data from AWS to GCP") {
     val moveTableName = "LINEITEM_FROM_PARQUET"
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
@@ -1196,10 +1196,12 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
   test("count super large table") {
     if (!skipBigDataTest) {
       val expectedCount = Int.MaxValue + 1L
-      val actualCount = getRowCount(
-        s"table(generator(rowcount => $expectedCount))",
-        thisConnectorOptionsNoTable
-      )
+      val actualCount = sparkSession.read
+        .format(SNOWFLAKE_SOURCE_NAME)
+        .options(thisConnectorOptionsNoTable)
+        .option("query", s"select 1 from table(generator(rowcount => $expectedCount))")
+        .load()
+        .count()
       assert(actualCount == expectedCount)
     }
   }
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
@@ -130,7 +130,7 @@ private[snowflake] case class SnowflakeRelation(
             sqlContext.getConf("spark.sql.shuffle.partitions", "200").toInt
           val emptyRow = Row.empty
           sqlContext.sparkContext
-            .parallelize(1L to numRows, parallelism)
+            .range(start = 0L, end = numRows, numSlices = parallelism)
             .map(_ => emptyRow)
         } else {
           throw new IllegalStateException("Could not read count from Snowflake")


src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -1193,6 +1193,17 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
     assert(count == LARGE_TABLE_ROW_COUNT)
   }
 
+  test("count super large table") {
```
sfc-gh-mrui, line 1196: @dohongdayi Thanks a lot for your PR on spark connector.
`getRowCount()` is a test function to run a query, It is not related to Spark Connector production code.
So, I don't think this test reproduced the issue.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -130,7 +130,7 @@ private[snowflake] case class SnowflakeRelation(
             sqlContext.getConf("spark.sql.shuffle.partitions", "200").toInt
           val emptyRow = Row.empty
           sqlContext.sparkContext
-            .parallelize(1L to numRows, parallelism)
+            .range(start = 0L, end = numRows, numSlices = parallelism)
```
sfc-gh-mrui, line 133: For `DataFrame.count()`, if it is pushdown to snowflake, the final return be a single Long value.
I don't understand why this change can affect the issue.

dohongdayi: @sfc-gh-mrui Thanks for your reply.
Yes, and here is building RDD with `result` number of empty rows, but `SparkContext.parallelize()` will fail to do it, for its first parameter `Seq[T]`'s `length()` only supports Int32

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
dohongdayi: Sure, please review again.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
dohongdayi: Query wouldn't be pushed down if spark version is lower than 3.1 and `requiredColumns` of `buildScan()` is empty, instead, it will get here.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
sfc-gh-mrui: This case can succeed without any fix.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
sfc-gh-mrui: I think the `count()` can be pushdown for  spark 3.1.
Please make sure you select the corresponding spark connector version for the spark version.
For example, for spark connector 2.9.1, we released binaries for spark 2.4, 3.0 and 3.1.

dohongdayi: Right, we are using Spark 3.0.2 and spark connector 2.9.1

dohongdayi: https://mvnrepository.com/artifact/net.snowflake/spark-snowflake_2.12/2.9.1-spark_3.0

wkeithvan:
Any progress on getting a fix for this through?

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
sfc-gh-mrui: @dohongdayi I see, the issue can only be reproduced if pushdown is disabled. Could you please update the test case to disable pushdown?
```
  test("count super large table") {
    if (!skipBigDataTest) {
      SnowflakeConnectorUtils.disablePushdownSession(sparkSession)
      val expectedCount = Int.MaxValue + 1L
      val actualCount = sparkSession.read
        .options(thisConnectorOptionsNoTable)
        .option("autopushdown", "false")
        .option("query", s"select 1 from table(generator(rowcount => $expectedCount))")
        .load()
        .count()
      assert(actualCount == expectedCount)
    }
  }
```

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
sfc-gh-mrui: @dohongdayi After explicitly disable pushdown for the test case, I can reproduce the issue. And I can ack this fix can fix that problem. After you update the test case. I can approve this PR.

sfc-gh-mrui (approved):
LGTM

sfc-gh-bli (approved):
LGTM

PR #394, opened 2021-11-10
SNOW-495389 Support to write empty DataFrame without CREATE TABLE permission if possible.

sfc-gh-mrui:
SNOW-495389 Support to write empty DataFrame without CREATE TABLE permission if possible.

Issue description
=============
Currently, when spark connector writes an empty DataFrame (without any partition), it executes a CREATE TABLE AS SELECT command. In most case, there is no problem because the user needs to have CREATE TABLE privilege.
But in below special case, the user actually doesn't need CREATE TABLE privilege.
1. truncate_table = on
2. usestagingtable = off
3. target table already exists.

Fixing
=====
The fix is to make the DataFrame write to go through the normal DataFrame process instead of executing CTAS for this special case. The original reason to use CTAS is that there is no file uploading happen, the driver can't get a stage prefix. It can be workaround by using a dummy prefix.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -173,6 +173,13 @@ object Parameters {
   val PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME: String = knownParam(
     "internal_skip_write_when_writing_empty_dataframe"
   )
+  // Internal option to use COPY command to write a DataFrame
+  // without any partitions. By default, it is 'true'.
+  // This option may be removed without any notice in any time.
+  // Introduced since Spark Connector 2.9.3
+  val PARAM_INTERNAL_USE_COPY_TO_WRITE_EMPTY_DATAFRAME: String = knownParam(
+    "internal_use_copy_to_write_empty_dataframe"
+  )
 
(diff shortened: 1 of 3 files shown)


sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -173,6 +173,13 @@ object Parameters {
   val PARAM_INTERNAL_SKIP_WRITE_WHEN_WRITING_EMPTY_DATAFRAME: String = knownParam(
     "internal_skip_write_when_writing_empty_dataframe"
   )
+  // Internal option to use COPY command to write a DataFrame
+  // without any partitions. By default, it is 'true'.
+  // This option may be removed without any notice in any time.
+  // Introduced since Spark Connector 2.9.3
+  val PARAM_INTERNAL_USE_COPY_TO_WRITE_EMPTY_DATAFRAME: String = knownParam(
```
sfc-gh-bli, line 180: I don't think we need this parameter to control it. do we?

sfc-gh-bli (approved):

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: Yes, this is a bug fixing, and the possibility to this the problem is low. I will remove this parameter.

Issue #386, opened 2021-10-25
Support for Spark 3.2


jpugliesi:
[Spark version 3.2 was released recently](https://spark.apache.org/news/spark-3-2-0-released.html) - this issue tracks updating `spark-snowflake` to support Spark 3.2

sfc-gh-mrui:
Spark connector 2.10.0 has been released which supports Spark 3.2.

PR #400, opened 2021-12-17
Build for spark 3.2.0

pdambrauskas:
Fixes: https://github.com/snowflakedb/spark-snowflake/issues/386

I'm not quite familiar with Scala and Spark internals, so I'm not sure if it is ok to leave `withNewChildrenInternal` without implementation.


sfc-gh-mrui:
@pdambrauskas Appreciated so much for your contribution! 
In general, it looks this PR is making the spark connector to compile with Spark 3.2 library.
As for Spark 3.2 support, it may need more changes on pushdown process.
The reason is that ,for each Spark version upgrade,
some interfaces may change, so the pushdown need to accommodate the changes.

pdambrauskas:
> @pdambrauskas Appreciated so much for your contribution! In general, it looks this PR is making the spark connector to compile with Spark 3.2 library. As for Spark 3.2 support, it may need more changes on pushdown process. The reason is that ,for each Spark version upgrade, some interfaces may change, so the pushdown need to accommodate the changes.

Hey, thanks for your response.
On this PR you can see, that there are some interface changes:
1. `Cast` object gets new property `ansiEnabled` Related change on spark: https://github.com/apache/spark/commit/6f51e37eb52f21b50c8d7b15c68bf9969fee3567
Since MiscStatement was not using tihis conf property from `SQLConf`, I assume, we can ignore it.

2. SubqueryExpression refactor: https://github.com/apache/spark/commit/806da9d6fae403f88aac42213a58923cf6c2cb05
Now it holds `joinCond` and `outerAttrs`. Plan itself did not change, so I assume no additional changes are needed?

3. withNewChildrenInternal on TreeNode: https://github.com/apache/spark/commit/0945baf90660a101ae0f86a39d4c91ca74ae5ee3
Not sure what should I do about this one. 
On solution description it states:
> It returns the same instance if the provided children are the same as its children, i.e., it preserves referential equality.

So I should probably return `this`, since children is overridden to `Nil`?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
```diff
@@ -30,14 +30,21 @@ object SnowflakeConnectorUtils {
...
+
   /**
     * Check Spark version, if Spark version matches SUPPORT_SPARK_VERSION enable PushDown,
     * otherwise disable it.
     */
   val SUPPORT_SPARK_VERSION = "3.1"
 
   def checkVersionAndEnablePushdown(session: SparkSession): Boolean =
-    if (session.version.startsWith(SUPPORT_SPARK_VERSION)) {
+    if (checkVersionHigherOrEqualTo(session.version, SUPPORT_SPARK_VERSION)) {
```
sfc-gh-mrui, line 47: The interface classes for different spark version may be different,
So one spark connector version, we release multiple binaries for different version.
1. Don't introduce `checkVersionHigherOrEqualTo()`
2. Change `val SUPPORT_SPARK_VERSION = "3.2"`

src/main/scala/net/snowflake/spark/snowflake/pushdowns/SnowflakePlan.scala
```diff
@@ -10,6 +10,8 @@ import org.apache.spark.sql.types.{StructField, StructType}
 case class SnowflakePlan(output: Seq[Attribute], rdd: RDD[InternalRow])
     extends SparkPlan {
 
+  override protected def withNewChildrenInternal(newChildren: IndexedSeq[SparkPlan]): SparkPlan = this
+
```
sfc-gh-mrui, line 14: Why `withNewChildrenInternal` is added?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
```diff
@@ -43,7 +43,7 @@ private[querygeneration] object MiscStatement {
     Option(expr match {
       case Alias(child: Expression, name: String) =>
         blockStatement(convertStatement(child, fields), name)
-      case Cast(child, t, _) =>
+      case Cast(child, t, _, _) =>
```
sfc-gh-mrui, line 46: For the interface change, we need to analysis:
1. What's the new parameter.
2. Can it be supported by snowflake corresponding function?

If it can be supported by snowflake, we will support it. Otherwise, we don't pushdown it.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/SnowflakePlan.scala
pdambrauskas: It is a new abstract function introduced in TreeNode, which is extended by this class. see more on my previous comment.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
pdambrauskas: I mentioned it in my previous comment it is 'ansiEnabled' parameter, which was previously accessible through SQLConf. It is not new, only the way it is accessed has changed: https://github.com/apache/spark/commit/6f51e37eb52f21b50c8d7b15c68bf9969fee3567 ansi support is expermental feature on spark, as I see it throws exceptions, when overflow on some numeric operations happens, and restricts some keywords: https://spark.apache.org/docs/3.2.0/sql-ref-ansi-compliance.html as I understand, snowflake supports ansi dialect by default, should we think of a way to change behaviour, when ansi is disabled? Return nulls instead of overflow exceptions, etc...?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
pdambrauskas: Got it! Done.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
sfc-gh-mrui: If this is the case, you may need to investigate more:
1. Does snowflake support `ansiEnabled = true` and `ansiEnabled = false`, if so, `ansiEnabled` should be passed to snowflake.
2. If no, does the snowflake CAST to be aligned with ansiEnabled or not, we may only pushdown for one case. For example, 
`case Cast(child, t, _, ansiEnabled) if ansiEnabled =>` 
or 
`case Cast(child, t, _, ansiEnabled) if !ansiEnabled =>`

src/main/scala/net/snowflake/spark/snowflake/pushdowns/SnowflakePlan.scala
sfc-gh-mrui: If this PR need to be merged into `master`, we have to figure out what's its use case and implement a reasonable one.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
```diff
@@ -104,7 +104,7 @@ private[querygeneration] object MiscStatement {
       case SortOrder(child, Descending, _, _) =>
         blockStatement(convertStatement(child, fields)) + "DESC"
 
-      case ScalarSubquery(subquery, _, _) =>
+      case ScalarSubquery(subquery, _, _, _) =>
```
sfc-gh-mrui, line 107: I saw your comments about `ScalarSubquery`. But, we can't `guess`. We have to make sure its usage.
If the new field is ignored, but it should not, it may cause wrong result. It is very difficult to diagnose this kind of problem.

sfc-gh-mrui: One more general comment is: Do you have a test case to test `Case` and `ScalarSubquery`?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
pdambrauskas: Thanks for taking time to answer.

In this case, default behaviour on Snowflakes side should be pushing down `TRY_CAST` (https://docs.snowflake.com/en/sql-reference/functions/try_cast.html)
Which returns null, when cast fails, and using simple `CAST`, when ansiEnabled

That would align with statement from spark documentation:
> When spark.sql.ansi.enabled is set to true, Spark SQL uses an ANSI compliant dialect instead of being Hive compliant. For example, Spark will throw an exception at runtime instead of returning null results if the inputs to a SQL operator/function are invalid.

Would you agree on that?

src/main/scala/net/snowflake/spark/snowflake/pushdowns/SnowflakePlan.scala
pdambrauskas: It is an optimization for `TreeNode#mapChildren`. This function doeas a check `containsChild.nonEmpty` which is allways `false` in case of `SnowflakePlan`, because there is:

```scala
override def children: Seq[SparkPlan] = Nil
```

so this function is never called, I just made it return `this`, as suggested in https://issues.apache.org/jira/browse/SPARK-34989. This can just throw an exception, since it never gets called for a childless node.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/MiscStatement.scala
pdambrauskas: Regarding `ScalarSubquery`, I looked into it one more time.
Before the update it was:
`ScalarSubquery(plan, children, exprId)`, now it is `ScalarSubquery(plan, outerAttrs, exportId, joinCond)`.
Previously both, `outerAttrs` and `joinCond` references were stored on `children`. It was done so that the same field was not used for different purposes and to avoid confusion and has no effect on LogicalPlan itself.

Regarding tests.
Since I did not implement any changes other than a bit of tuning to updated interfaces, I was relying on existing tests, I saw there are some tests covering `CAST` and sub-queries.

If you agree with my suggestion about `CAST`(when ansi enabled) and `TRY_CAST` (by default), which I mentioned on my previous comment, I'll add a testcase for that.

Thanks again, for responding, and sorry for my `guesses`, let me know if my last comments make any sense, and we can move forward with this.

liebstein: Hey. Any plans to add this for the next 2.9.x release?

liebstein:
Hey. Any plans to add this for the next 2.9.x release?

sfc-gh-mrui:
@pdambrauskas The Spark Connector team has decided to support it in next spark connector release. I will work from your contribution. Appreciate for your contribution on the Spark 3.2 support!

sfc-gh-mrui:
@pdambrauskas I have merged 2 PRs to support Spark 3.2. If you need it, you can build from master branch.
https://github.com/snowflakedb/spark-snowflake/pull/408
https://github.com/snowflakedb/spark-snowflake/pull/407

liebstein:
Thanks for the update

PR #407, opened 2022-02-02
SNOW-540761 Support Spark 3.2 Scala 2.13

sfc-gh-mrui:


sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
```diff
@@ -18,7 +18,7 @@
 package net.snowflake.spark.snowflake
 
 import java.sql.{Date, Timestamp}
-
+import scala.language.postfixOps
```
sfc-gh-bli, line 21: what is it?

sfc-gh-bli: Is it supported by both Scala 2.12 and 2.13?

sfc-gh-bli (approved):

src/main/scala/net/snowflake/spark/snowflake/FilterPushdown.scala
sfc-gh-mrui: Scala 2.13 needs to import `scala.language.postfixOps` explicitly if postfix operator is used such as `!`.
```
    // clean up download dir
    import sys.process._
    "rm -fr /tmp/test_move_data" !
```

sfc-gh-mrui: and yet. It is supported for both.

PR #408, opened 2022-02-05
SNOW-540759 Spark 3.2 Scala 2.12 support

sfc-gh-mrui:
SNOW-540759 Spark 3.2 Scala 2.12 support


src/it/scala/net/snowflake/spark/snowflake/ColumnNameCaseSuite.scala
```diff
@@ -174,7 +174,16 @@ class ColumnNameCaseSuite extends IntegrationSuiteBase {
          |"SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0" ) AS "SUBQUERY_1" WHERE
          |( ( "SUBQUERY_1"."SUBQUERY_1_COL_2" IS NOT NULL ) AND ( "SUBQUERY_1".
          |"SUBQUERY_1_COL_2" = 1 ) )
-         |""".stripMargin.replaceAll("\\s", ""))
+         |""".stripMargin.replaceAll("\\s", "")) ||
+        Utils.getLastSelect
```
sfc-gh-mrui, line 178: The spark plan in Spark 3.2 is different comparing with Spark 3.1. So the pushdown query is different. 
This affects a lot of test case.

We want the test case to support both 3.2 and 3.1/3.0. So, I will update the affected test case.
If the actual query is equals to either version, we think the test case to be true.

1. This test case checked the query directly, so I update it directly.
2. In many test cases, `testPushdown()` is used to check pushdown result and query.
    I will create a new test method: `testPushdownMultiplefQueries(queries: Seq[String],...)`
    which will check the last query to be equal to any query in `queries`.

.github/workflows/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.5.5
```
sfc-gh-bli, line 1: Does it work?

sfc-gh-bli (approved):

.github/workflows/project/build.properties
sfc-gh-mrui: This is needed for Scala 2.13 support, I will remove this change in this PR.

src/it/scala/net/snowflake/spark/snowflake/PushdownEnhancement02.scala
```diff
@@ -509,9 +524,48 @@ class PushdownEnhancement02 extends IntegrationSuiteBase {
...
+         | "SUBQUERY_2_COL_2" FROM ( SELECT ( "SUBQUERY_0"."STATE" )
+         | AS "SUBQUERY_1_COL_0" , ( "SUBQUERY_0"."BUSHELS_PRODUCED" ) AS
+         | "SUBQUERY_1_COL_1" FROM ( SELECT * FROM ( $test_table_rank )
+         | AS "SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0" ) AS "SUBQUERY_1"
+         |""".stripMargin
     )
+    testPushdownMultiplefQueries(expectedQueries, resultDF, expectedResult)
+  }
+
+  test("literal Date/Timestamp") {
```
sfc-gh-mrui, line 541: This is a new test case to test code change in `BasicStatement.scala`

src/it/scala/org/apache/spark/sql/snowflake/SFTestData.scala
```diff
@@ -12,84 +12,114 @@ trait SFTestData extends SQLTestData {
 
   import org.apache.spark.sql.test.SQLTestData.CourseSales
 
-  override protected lazy val testData: DataFrame =
-    spark.createDataFrame((1 to 100).map(i => TestData(i, i.toString)))
+  override protected lazy val testData: DataFrame = {
```
sfc-gh-mrui, line 15: In some new test case for Spark 3.2, it may use the `view` for the test DataFrame.
To make sure that kind of test case work, We also need to create a view after we convert a local relation to a Snowflake table.

sfc-gh-ema (approved):

PR #413, opened 2022-02-15
SNOW-528863 Fix an issue that string "null" is regarded as NULL

sfc-gh-mrui:
SNOW-528863 Fix an issue that literal string "null" is regarded as NULL for pushdown.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
@@ -84,7 +84,7 @@ private[querygeneration] object BasicStatement {
       case l: Literal =>
         l.dataType match {
           case StringType =>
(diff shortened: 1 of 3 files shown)


src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/BasicStatement.scala
```diff
@@ -84,7 +84,7 @@ private[querygeneration] object BasicStatement {
       case l: Literal =>
         l.dataType match {
           case StringType =>
-            if (l == null || l.toString() == "null") {
+            if (l.value == null) {
```
sfc-gh-mrui, line 87: Literal String `"null"` is regarded as NULL because `l.toString()` will be `"null"` for both below case:
1. `l` is a StringType and its value is NULL
2. `l` is a StringType and its value is string `"null"`

sfc-gh-bli (approved):

PR #431, opened 2022-06-16
SNOW-614357 Support JDBC datatype TIMESTAMP_WITH_TIMEZONE

sfc-gh-mrui:
JDBC only changes the returned datatype, it doesn't change how to get the value from ResultSet.
So SC only needs to support the new type when resolving the schema.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
@@ -461,6 +461,7 @@ private[snowflake] class JDBCWrapper {
       case java.sql.Types.STRUCT    => StringType // Snowflake-todo: ?
       case java.sql.Types.TIME      => StringType
       case java.sql.Types.TIMESTAMP => TimestampType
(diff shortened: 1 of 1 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -461,6 +461,7 @@ private[snowflake] class JDBCWrapper {
       case java.sql.Types.STRUCT    => StringType // Snowflake-todo: ?
       case java.sql.Types.TIME      => StringType
       case java.sql.Types.TIMESTAMP => TimestampType
+      case java.sql.Types.TIMESTAMP_WITH_TIMEZONE => TimestampType
```
sfc-gh-mrui, line 464: JDBC only changes the returned datatype, it doesn't change how to get the value from ResultSet.
So SC only needs to support the new type when resolving the schema.

The spark connector regression test detect the problem. It is not necessary to extra test.

sfc-gh-bli (approved):
LGTM

sfc-gh-sshankar (approved):
Approval

PR #440, opened 2022-07-19
SNOW-521177 Don't use new connection to check table existence when saving DF

sfc-gh-mrui:
SNOW-521177 Don't use new connection to check table existence when saving DF

Issue description
=============
The customer uses OAuth authentication to connect to snowflake and the token will expire soon after creating the connection. When writing a DataFrame to a table, SC uses a new connection to check the target table existence. The issue is that the token has expired when SC creates the new connection.

Reason to introduce the extra connection
===============================
The extra connection for table existence check is to resolve below case.
1. Suppose user sets `sfSchema = schema1` and will write DF to table `t1`.
2. `t1` doesn't exist in `schema1`, but it exists in schema `public`. The table existence check will return `true` because  unqualified object names are resolved through a search path and by default, the search path is `schema1, public`. For details refer to https://docs.snowflake.com/en/sql-reference/name-resolution.html#queries .
3. To resolve this issue, SC creates a new connection and executes `"alter session set search_path='$current'"` to make it only search in current schema.
4. The reason to introduce a new connection is to avoid the `alter session` command affect other following queries such as `postAction`.

Reference:
The PR to introduce extra connection: https://github.com/snowflakedb/spark-snowflake/pull/327
The Jira for the issue: https://snowflakecomputing.atlassian.net/browse/SNOW-262080

This fix proposal
=============
In this new proposal, we don't introduce new connection, and check the table existence with fully qualified name in current connection.
1. If the target table name is fully qualified name, just use it.
2. If it is <schema_name>.<table_name>, SC will add sfDataBase to it. e.g. <sfDatabase>. <schema_name>.<table_name>
3. If it is <table_name>, SC will add sfDatabase and sfSchema, e.g.  <sfDatabase>. <sfSchema>.<table_name>

NOTE:
1. This change is protected by an internal parameter
2. sfDatabase and sfSchema are Obligatory options.
3. We choose to use sfDatabase and sfSchema instead of executing a query to get currentDatabase and currentSchema because it is not necessary. The only possibility of current schema/database to be different with sfSchema/sfDatabase is that they executes `use database/schema` in preActions. This sounds like not a right way to use SC. If they do needs to write a table in different schema, they should use schema name in the target table.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
@@ -725,39 +725,6 @@ object Utils {
     // scalastyle:off println
   }
 
-  // Find the first un-quoted char '.', return -1 if not found
-  private def findFirstUnquotedDotIndex(data: String): Int = {
-    var quoteCount = 0
-    for (i <- 0 until data.length) {
-      if (data(i) == '"') {
-        quoteCount = quoteCount + 1
-      } else if (data(i) == '.') {
-        if (quoteCount % 2 == 0) {
-          return i
-        }
-      }
-    }
-    -1
-  }
-
-  // Split object name with unquoted dot.
-  private[snowflake] def splitNameByDot(name: String): Seq[String] = {
-    val result = new mutable.ArrayBuffer[String]
-    var data = name
-    var done = false
-    while (!done) {
-      val dotIndex = findFirstUnquotedDotIndex(data)
-      if (dotIndex == -1) {
-        result.append(data)
-        done = true
-      } else {
-        result.append(data.substring(0, dotIndex))
-        data = data.substring(dotIndex + 1)
-      }
-    }
-    result.toSeq
-  }
-
   // Adjust table name by adding database or schema name for table existence check.
   private[snowflake] def getTableNameForExistenceCheck(database: String,
                                                        schema: String,
@@ -766,16 +733,16 @@ object Utils {
     val quotedIdPattern = """("([^"]|"")+")"""
     val idPattern = s"($unQuotedIdPattern|$quotedIdPattern)"
 
-    val nameParts = splitNameByDot(name)
-    //     identifier
-    //     identifier.identifier
-    //     identifier.identifier.identifier
-    //     identifier..identifier // It's equals to identifier.PUBLIC.identifier
-    if (nameParts.length == 1 && name.matches(idPattern)) {
+    // scalastyle:off
+    // Only add the database/schema name for <id> and <id>.<id>
+    // If the name is fully qualified name or invalid name, don't need to adjust it.
+    // NOTE: <id>..<id> is equals to <id>.PUBLIC.<id> which is a fully qualified name.
+    // refer to https://docs.snowflake.com/en/sql-reference/name-resolution.html#resolution-when-schema-omitted-double-dot-notation
+    // scalastyle:off
+    if (name.matches(idPattern)) {
       // Add database and schema name
       s"${ensureQuoted(database)}.${ensureQuoted(schema)}.$name"
-    } else if (nameParts.length == 2 && nameParts.head.matches(idPattern)
-      && nameParts.last.matches(idPattern)) {
+    } else if (name.matches(s"$idPattern\\.$idPattern")) {
       // Add database name
       s"${ensureQuoted(database)}.$name"
     } else {
--- a/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Utils.scala
@@ -724,4 +724,64 @@ object Utils {
     }
     // scalastyle:off println
   }
+
+  // Find the first un-quoted char '.', return -1 if not found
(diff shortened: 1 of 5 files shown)


sfc-gh-ema:
@sfc-gh-mrui Why don't we just use a fully qualified name to perform the existence check? Why set a session parameter?

sfc-gh-mrui:
> @sfc-gh-mrui Why don't we just use a fully qualified name to perform the existence check? Why set a session parameter?

It's a potential solution to use `fully qualified name`, the reason we don't use that is that it is kind of complex because the target name may include schema name or database name. I can investigate this proposal again. 

By the way, `SEARCH_PATH` is a `SESSION_LINEAGE` parameter, so we can't set is as statement parameter.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,64 @@ object Utils {
...
+        done = true
+      } else {
+        result.append(data.substring(0, dotIndex))
+        data = data.substring(dotIndex + 1)
+      }
+    }
+    result
+  }
+
+  // Adjust table name by adding database or schema name for table existence check.
```
sfc-gh-mrui, line 761: 1. The users should input a valid object name.
2. If it is invalid, this function just return the original name. it will fail in the query.
3. Totally, there are 4 valid format for the object name. SC only adjust 2 of them because the other 2 are fully qualified names. It is not necessary to adjust them.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,64 @@ object Utils {
...
+    }
+    result.toSeq
+  }
+
+  // Adjust table name by adding database or schema name for table existence check.
+  private[snowflake] def getTableNameForExistenceCheck(database: String,
+                                                       schema: String,
+                                                       name: String): String = {
+    val unQuotedIdPattern = """([a-zA-Z_][\w$]*)"""
+    val quotedIdPattern = """("([^"]|"")+")"""
```
sfc-gh-bli, line 766: I like this

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,64 @@ object Utils {
...
+    val idPattern = s"($unQuotedIdPattern|$quotedIdPattern)"
+
+    val nameParts = splitNameByDot(name)
+    //     identifier
+    //     identifier.identifier
+    //     identifier.identifier.identifier
+    //     identifier..identifier // It's equals to identifier.PUBLIC.identifier
```
sfc-gh-bli, line 773: Is it always be `public`? can it be other default schema?

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,64 @@ object Utils {
...
+
+  // Adjust table name by adding database or schema name for table existence check.
+  private[snowflake] def getTableNameForExistenceCheck(database: String,
+                                                       schema: String,
+                                                       name: String): String = {
+    val unQuotedIdPattern = """([a-zA-Z_][\w$]*)"""
+    val quotedIdPattern = """("([^"]|"")+")"""
+    val idPattern = s"($unQuotedIdPattern|$quotedIdPattern)"
+
+    val nameParts = splitNameByDot(name)
```
sfc-gh-bli, line 769: since you have regex here anyway, the `splitNameByDot` function is not good.

sfc-gh-bli (approved):
I am ok with this solution.

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,64 @@ object Utils {
     }
     // scalastyle:off println
   }
+
+  // Find the first un-quoted char '.', return -1 if not found
```
sfc-gh-ema, line 728: Since you split this out as a separate function, should you have a couple of unit tests for it?

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -724,4 +724,31 @@ object Utils {
...
+    val quotedIdPattern = """("([^"]|"")+")"""
+    val idPattern = s"($unQuotedIdPattern|$quotedIdPattern)"
+
+    // scalastyle:off
+    // Only add the database/schema name for <id> and <id>.<id>
+    // If the name is fully qualified name or invalid name, don't need to adjust it.
+    // NOTE: <id>..<id> is equals to <id>.PUBLIC.<id> which is a fully qualified name.
+    // refer to https://docs.snowflake.com/en/sql-reference/name-resolution.html#resolution-when-schema-omitted-double-dot-notation
+    // scalastyle:off
+    if (name.matches(idPattern)) {
```
sfc-gh-bli, line 742: good

sfc-gh-ema:
LGTM.

PR #442, opened 2022-07-22
SNOW-548245 (Test Only) Add unit test for unionByName pushdown

sfc-gh-mrui:
Customer requests to support UNION BY NAME pushdown.
I found that Spark has optimized the UNION BY NAME as UNION in the spark plan. So, it actually has been supported without any SC enhancement required.

This PR is to add some unit test for `unionByName` pushdown.


sfc-gh-ema (approved):

sfc-gh-bli (approved):
LGTM

PR #444, opened 2022-08-01
SNOW-638510 Add some fields to SPARK_CLIENT_INFO

sfc-gh-mrui:
These PR added 3 items:
1. Add language name to SPARK_CLIENT_INFO
2. Add spark configuration names if they are not in the white-list. ("N/A" is set in the message for the value)
3. Add SPARK_APPLICATION_ID to all telemetry messages.
4. Introduce `TelemetryMessageSender` for testing purpose.

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
@@ -11,8 +11,10 @@ import org.apache.spark.sql.types._
 import scala.collection.mutable
 import scala.util.Random
 
-private[snowflake] class FakeTelemetryMessageSender(buffer: mutable.ArrayBuffer[ObjectNode])
+private[snowflake] class MockTelemetryMessageSender(buffer: mutable.ArrayBuffer[ObjectNode])
   extends TelemetryMessageSender {
+  // the telemetry messages are appended into the buffer instead of sending to snowflake,
+  // So that the test case can check the message to be sent.
   override def send(telemetry: Telemetry, logs: List[(ObjectNode, Long)]): Unit = {
     logs.foreach {
       case (log, _) => buffer.append(log)
@@ -55,7 +57,7 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
     // Enable dummy sending telemetry message.
     val messageBuffer = mutable.ArrayBuffer[ObjectNode]()
     val oldSender = SnowflakeTelemetry.setTelemetryMessageSenderForTest(
-      new FakeTelemetryMessageSender(messageBuffer))
+      new MockTelemetryMessageSender(messageBuffer))
     try {
       // A basis dataframe read
       val df1 = sparkSession.read
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
@@ -38,7 +38,7 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
     assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("Scala"))
     // Manually set pyspark
     sparkConf.set("spark.pyspark.python", "/usr/bin/python3")
-    assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("python3"))
+    assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("Python"))
     // Manually set R language
     sparkConf.set("spark.r.command", "R")
     assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("R"))
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
@@ -1,0 +1,86 @@
+package net.snowflake.spark.snowflake
+
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.ObjectMapper
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode
+import net.snowflake.spark.snowflake.SnowflakeTelemetry.mapper
+import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
+import org.apache.spark.SparkEnv
+import org.apache.spark.sql.{Row, SaveMode}
+import org.apache.spark.sql.types._
+
+import scala.collection.mutable
+import scala.util.Random
+
+class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
+
+  private val test_table: String = s"test_table_1_$randomSuffix"
+  private val test_table2: String = s"test_table_2_$randomSuffix"
+
+  private val mapper = new ObjectMapper
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+
+    SnowflakeConnectorUtils.enablePushdownSession(sparkSession)
+  }
+
+  test("unit test: SnowflakeTelemetry.detectSparkLanguage") {
+    val sparkConf = SparkEnv.get.conf.clone()
+    assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("Scala"))
+    // Manually set pyspark
+    sparkConf.set("spark.pyspark.python", "/usr/bin/python3")
+    assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("python3"))
+    // Manually set R language
+    sparkConf.set("spark.r.command", "R")
+    assert(SnowflakeTelemetry.detectSparkLanguage(sparkConf).equals("R"))
+  }
+
+  test("unit test: SnowflakeTelemetry.addCommonFields") {
+    val metric: ObjectNode = mapper.createObjectNode()
+    SnowflakeTelemetry.addCommonFields(metric)
+    assert(metric.size() == 1 &&
+      metric.get(TelemetryFieldNames.SPARK_APPLICATION_ID).asText().startsWith("local-"))
+  }
+
+  test("IT test: common fields are added") {
+    try {
+      // Enable dummy sending telemetry message.
+      val fakeMessageSender = mutable.ArrayBuffer[ObjectNode]()
+      SnowflakeTelemetry.setFakeMessageSender(fakeMessageSender)
+
+      // A basis dataframe read
+      val df1 = sparkSession.read
+        .format(SNOWFLAKE_SOURCE_NAME)
+        .options(connectorOptionsNoTable)
+        .option("query", "select current_timestamp()")
+        .load()
+      df1.collect()
+
+      fakeMessageSender.foreach { x =>
+        val typeName = x.get("type").asText()
+        val source = x.get("source").asText()
+        assert(source.equals("spark_connector"))
+        val data = x.get("data")
+        assert(data.get(TelemetryFieldNames.SPARK_APPLICATION_ID).asText().startsWith("local-"))
+        if (typeName.equals("spark_client_info")) {
+          // Spark language is added for spark_client_info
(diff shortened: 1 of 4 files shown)


src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -0,0 +1,86 @@
...
+    SnowflakeTelemetry.addCommonFields(metric)
+    assert(metric.size() == 1 &&
+      metric.get(TelemetryFieldNames.SPARK_APPLICATION_ID).asText().startsWith("local-"))
+  }
+
+  test("IT test: common fields are added") {
+    try {
+      // Enable dummy sending telemetry message.
+      val fakeMessageSender = mutable.ArrayBuffer[ObjectNode]()
+      SnowflakeTelemetry.setFakeMessageSender(fakeMessageSender)
```
sfc-gh-mrui, line 49: If fake message sender is set, the telemetry message is appended into the ArrayBuffer instead of sending to snowflake, and the test can check the message to be sent.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -34,6 +35,20 @@ object SnowflakeTelemetry {
 
   private[snowflake] var output: ObjectNode = _
 
+  private var testFakeMessagesSender: Option[mutable.ArrayBuffer[ObjectNode]] = None
+
+  private[snowflake] def setFakeMessageSender(fakeSender: mutable.ArrayBuffer[ObjectNode]): Unit =
+    testFakeMessagesSender = Option(fakeSender)
+
+  private lazy val sparkApplicationId: String =
```
sfc-gh-mrui, line 43: sparkApplicationId is added to all telemetry messages. So set it as lazy val.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -145,16 +161,23 @@ object SnowflakeTelemetry {
...
+                          |Send Telemetry
+                          |timestamp:$timestamp
+                          |log:${log.toString}"
            """.stripMargin)
-        telemetry.asInstanceOf[TelemetryClient].addLogToBatch(log, timestamp)
+          telemetry.asInstanceOf[TelemetryClient].addLogToBatch(log, timestamp)
+      }
+      telemetry.sendBatchAsync()
+    } else {
+      // This is for test only
```
sfc-gh-mrui, line 176: If fake message sender is set, the telemetry message is appended into the ArrayBuffer instead of sending to snowflake, and the test can check the message to be sent.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -273,18 +298,19 @@ object SnowflakeTelemetry {
...
-      sparkOptions.foreach(
-        optionName => {
-          if (sparkConf.contains(optionName)) {
-            sparkMetric.put(optionName, sparkConf.get(optionName))
+      sparkConf.getAll.foreach {
+        case (key, value) =>
+          if (sparkOptions.contains(key)) {
+            sparkMetric.put(key, value)
+          } else {
+            sparkMetric.put(key, "N/A")
```
sfc-gh-mrui, line 311: For the spark configuration which are not in the white-list, set value "N/A".

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -366,6 +392,22 @@ object SnowflakeTelemetry {
     }
     metric
   }
+
+  private[snowflake] def detectSparkLanguage(sparkConf: SparkConf): String = {
```
sfc-gh-mrui, line 396: Detect the language based on spark options.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -366,6 +392,22 @@ object SnowflakeTelemetry {
...
+    if (sparkConf.contains("spark.r.command")
+      || sparkConf.contains("spark.r.driver.command")
+      || sparkConf.contains("spark.r.shell.command")) {
+      "R"
+    } else if (sparkConf.contains("spark.pyspark.python")) {
+      sparkConf.get("spark.pyspark.python").split("/").last
```
sfc-gh-mrui, line 402: @sfc-gh-ema `sparkConf.get("spark.pyspark.python").split("/").last` will get the binary name. It may include the version number, but it is not reliable. What about we just use "Python"?

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -0,0 +1,95 @@
...
+import net.snowflake.client.jdbc.telemetry.Telemetry
+import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
+import org.apache.spark.SparkEnv
+import org.apache.spark.sql.{Row, SaveMode}
+import org.apache.spark.sql.types._
+
+import scala.collection.mutable
+import scala.util.Random
+
+private[snowflake] class FakeTelemetryMessageSender(buffer: mutable.ArrayBuffer[ObjectNode])
```
sfc-gh-mrui, line 14: With the fake `TelemetryMessageSender`, the telemetry messages are appended into the ArrayBuffer instead of sending to snowflake, and the test can check the message to be sent.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -145,16 +163,7 @@ object SnowflakeTelemetry {
...
-      case (log, timestamp) =>
-        logger.debug(s"""
-             |Send Telemetry
-             |timestamp:$timestamp
-             |log:${log.toString}"
-           """.stripMargin)
-        telemetry.asInstanceOf[TelemetryClient].addLogToBatch(log, timestamp)
-    }
-    telemetry.sendBatchAsync()
+    telemetryMessageSender.send(telemetry, curLogs)
```
sfc-gh-mrui, line 166: The implementation is moved to `RealTelemetryMessageSender.send()`

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -550,3 +579,25 @@ object TelemetryOOBTags {
   val CTX_PROTOCAL = "ctx_protocol"
   val CTX_USER = "ctx_user"
 }
+
+private[snowflake] trait TelemetryMessageSender {
+  def send(telemetry: Telemetry, logs: List[(ObjectNode, Long)]): Unit
+}
+
+private final class RealTelemetryMessageSender extends TelemetryMessageSender {
```
sfc-gh-bli, line 587: I don't like this name.

src/main/scala/net/snowflake/spark/snowflake/streaming/SnowflakeSink.scala
```diff
@@ -125,6 +125,7 @@ class SnowflakeSink(sqlContext: SQLContext,
         val time = System.currentTimeMillis()
         metric.put(END_TIME, time)
         metric.get(LOAD_RATE).asInstanceOf[ObjectNode].put(END_TIME, time)
+        SnowflakeTelemetry.addCommonFields(metric)
```
sfc-gh-bli, line 128: do we still maintain this stream connector?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui: any other idea? What about `SnowflakeTelemetryMessageSender` ?

src/main/scala/net/snowflake/spark/snowflake/streaming/SnowflakeSink.scala
sfc-gh-mrui: We don't officially support new features. It's not necessary to maintain it. Since it is some lines low risk change, it is better to maintain it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -366,6 +378,22 @@ object SnowflakeTelemetry {
...
     metric
   }
+
+  private[snowflake] def detectSparkLanguage(sparkConf: SparkConf): String = {
+    if (sparkConf.contains("spark.r.command")
+      || sparkConf.contains("spark.r.driver.command")
+      || sparkConf.contains("spark.r.shell.command")) {
+      "R"
+    } else if (sparkConf.contains("spark.pyspark.python")) {
+      "Python"
```
sfc-gh-mrui, line 388: @sfc-gh-ema `sparkConf.get("spark.pyspark.python").split("/").last` will get the binary name. It may include the version number, but it is not reliable. Since we can't detect the exact version, I  think we should use "Python"?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -34,6 +33,32 @@ object SnowflakeTelemetry {
...
 
+  private var telemetryMessageSender: TelemetryMessageSender = new SnowflakeTelemetryMessageSender
+
+  private[snowflake] def setTelemetryMessageSenderForTest(sender: TelemetryMessageSender)
+  : TelemetryMessageSender = {
+    val oldSender = telemetryMessageSender
+    telemetryMessageSender = sender
+    oldSender
+  }
+
```
sfc-gh-mrui, line 44: Use `cachedSparkApplicationId` + `getSparkApplicationId` instead of `lazy val` to avoid caching an invalid spark application id.
The potential use case is:
Users use `Util.runQuery()` to execute a query before creating SparkSession. `runQuery()` creates a connection so it may not get a valid spark application id.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
sfc-gh-bli: how about `MockTelemetryMessageSender`?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-bli: much better

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
sfc-gh-mrui: Will update to use `MockTelemetryMessageSender` in next PR update.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -366,6 +386,22 @@ object SnowflakeTelemetry {
...
+  private[snowflake] def detectSparkLanguage(sparkConf: SparkConf): String = {
+    if (sparkConf.contains("spark.r.command")
+      || sparkConf.contains("spark.r.driver.command")
+      || sparkConf.contains("spark.r.shell.command")) {
+      "R"
+    } else if (sparkConf.contains("spark.pyspark.python")) {
+      "Python"
+    } else {
+      "Scala"
```
sfc-gh-ema, line 398: nit: Is there some way to distinguish between Scala and Java? If not, should this be JVM or some combined label?

sfc-gh-mrui: Scala or Java is from development language perspective. The language detection is at RunTime.  I don't think we can distinguish them without some special Spark configuration. `JVM` sounds more generic. But I prefer to use `Scala` because Scala is well accepted in Spark world.

sfc-gh-bli (approved):
LGTM

sfc-gh-ema (approved):

PR #445, opened 2022-08-03
SNOW-637639 Add API to Utils to return the last read/write Query ID

sfc-gh-mrui:
SNOW-637639 Add API to Utils to return the last read/write Query ID


src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
```diff
@@ -197,6 +197,16 @@ trait IntegrationEnv
     val _ = conn.createStatement.executeQuery(query)
   }
 
+  def getQueryTextFromHistory(queryId: String): String = {
+    val rs = Utils.runQuery(connectorOptionsNoTable,
+      s"select query_text from table(information_schema.QUERY_HISTORY_BY_USER()) " +
+        s"where QUERY_ID ='$queryId' limit 1")
```
sfc-gh-ema, line 203: Do we need `limit 1`? Is it possible for there to be multiple rows for the same query id?

sfc-gh-ema (approved):

src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
sfc-gh-mrui: I think the query ID is unique. The `limit 1` is to make the query faster.

sfc-gh-bli (approved):
LGTM

PR #448, opened 2022-08-05
SNOW-640144 Add new telemetry message for spark plan statistic

sfc-gh-mrui:
SNOW-640144 Add new telemetry message for spark plan statistic.

A sample message for `read from snowflake, and write to snowflake` is:
```
{
  "type" : "spark_plan_statistic",
  "source" : "spark_connector",
  "data" : {
    "statistic_info" : [ 
      "LogicalRelation:SnowflakeRelation",
      "SaveIntoDataSourceCommand:DefaultSource",
      "org.apache.spark.sql.catalyst.plans.logical.ReturnAnswer",
      "org.apache.spark.sql.catalyst.expressions.AttributeReference"
      ],
    "spark_application_id" : "local-1659985199387"
  }
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,6 +56,8 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
       "org.scalatest" %% "scalatest" % "3.1.1" % Test,
       "org.mockito" % "mockito-core" % "1.10.19" % Test,
       "org.apache.commons" % "commons-lang3" % "3.5" % "provided",
+      // For test to read/write from postgresql
+      "org.postgresql" % "postgresql" % "42.3.6" % Test,
       // Below is for Spark Streaming from Kafka test only
       // "org.apache.spark" %% "spark-sql-kafka-0-10" % "2.4.0",
       "org.apache.spark" %% "spark-core" % testSparkVersion % "provided, test",
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
@@ -1,15 +1,17 @@
 package net.snowflake.spark.snowflake
 
+import java.nio.file.Files
+
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
 import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode
 import net.snowflake.client.jdbc.telemetry.Telemetry
 import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
 import org.apache.spark.SparkEnv
-import org.apache.spark.sql.{Row, SaveMode}
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.SaveMode
 
 import scala.collection.mutable
-import scala.util.Random
+import scala.reflect.io.Directory
 
 private[snowflake] class MockTelemetryMessageSender(buffer: mutable.ArrayBuffer[ObjectNode])
   extends TelemetryMessageSender {
@@ -28,6 +30,14 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
   private val test_table2: String = s"test_table_2_$randomSuffix"
 
   private val mapper = new ObjectMapper
+
+  private def nodeContains(node: JsonNode, value: String): Boolean = {
+    val iterator = node.iterator()
+    while (iterator.hasNext) {
+      if (iterator.next().asText().equals(value)) return true
+    }
+    false
+  }
 
   override def beforeAll(): Unit = {
     super.beforeAll()
@@ -110,6 +120,7 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
         .format(SNOWFLAKE_SOURCE_NAME)
         .options(connectorOptionsNoTable)
         .option("dbtable", test_table)
+        .mode(SaveMode.Overwrite)
         .save()
 
       // Check SPARK_EGRESS to be sent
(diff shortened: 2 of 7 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -105,6 +105,28 @@ object SnowflakeTelemetry {
     }
   }
 
+  private val MAX_CACHED_SPARK_PLAN_STATISTIC_COUNT = 1000
+
+  private[snowflake] def addSparkPlanStatistic(statistic: Set[String]): Unit = {
+    if (statistic.nonEmpty) {
+      this.synchronized {
+        if (logs.length >= MAX_CACHED_SPARK_PLAN_STATISTIC_COUNT) {
```
sfc-gh-mrui, line 113: If the pushdown is enabled, the spark plan statistic is added to `SnowflakeTelemetry.logs`, but it will be sent if any other kinds of messages is sent to snowflake lately.
In worst case, the pushdown is enabled, but the user doesn't not read/write from/to snowflake, all the spark plan statistic may be cached. To avoid OOM, SC cache at most 1000 messages.

src/main/scala/net/snowflake/spark/snowflake/pushdowns/querygeneration/QueryBuilder.scala
```diff
@@ -131,6 +129,40 @@ private[querygeneration] class QueryBuilder(plan: LogicalPlan) {
     }
   }
 
+  def analyzeSparkPlan(plan: LogicalPlan): Unit = try {
+    plan match {
+      case _: ReturnAnswer =>
+        val statisticSet = mutable.HashSet[String]()
```
sfc-gh-mrui, line 135: Spark may attempt to pushdown sub-plan, so we only handle the plan with `ReturnAnswer` on the top.

src/main/scala/org/apache/spark/sql/snowflake/SnowflakeSparkUtils.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.spark.sql.execution.datasources.jdbc.{JDBCRelation, JdbcRelationProvider}
+
+object SnowflakeSparkUtils {
+
+  private def getClassName(obj: Any): String = {
+    val shortClassNames = Seq("net.snowflake.spark.snowflake.SnowflakeRelation",
+      "net.snowflake.spark.snowflake.DefaultSource",
+      "org.apache.spark.sql.execution.datasources.jdbc.JdbcRelationProvider",
+      "org.apache.spark.sql.execution.datasources.jdbc.JDBCRelation")
+    val className = obj.getClass.getName
```
sfc-gh-mrui, line 13: We mostly query the spark plan statistic with these 4 classes, so use short name for them.

src/main/scala/org/apache/spark/sql/snowflake/SnowflakeSparkUtils.scala
```diff
@@ -0,0 +1,49 @@
+package org.apache.spark.sql.snowflake
+
+import org.apache.spark.sql.execution.datasources.{LogicalRelation, SaveIntoDataSourceCommand}
+import org.apache.spark.sql.execution.datasources.jdbc.{JDBCRelation, JdbcRelationProvider}
+
+object SnowflakeSparkUtils {
```
sfc-gh-mrui, line 6: `org.apache.spark.sql.JDBCRelation` is package private. So introduce `org.apache.spark.sql.snowflake.SnowflakeSparkUtils` to generate the class name.

build.sbt
```diff
@@ -56,6 +56,8 @@ lazy val root = project.withId("spark-snowflake").in(file("."))
       "org.scalatest" %% "scalatest" % "3.1.1" % Test,
       "org.mockito" % "mockito-core" % "1.10.19" % Test,
       "org.apache.commons" % "commons-lang3" % "3.5" % "provided",
+      // For test to read/write from postgresql
+      "org.postgresql" % "postgresql" % "42.3.6" % Test,
```
sfc-gh-mrui, line 60: Add `postgresql` as Test library because I want to add `"unit test: SnowflakeSparkUtils.getNameForLogicalPlanOrExpression for JDBCRelation"`

src/it/scala/org/apache/spark/sql/SnowflakeSparkUtilsSuite.scala
```diff
@@ -0,0 +1,69 @@
+package org.apache.spark.sql
+
+import org.apache.spark.sql.execution.datasources.{LogicalRelation, SaveIntoDataSourceCommand}
+import org.apache.spark.sql.execution.datasources.jdbc.{JDBCOptions, JDBCRelation, JdbcRelationProvider}
+import org.apache.spark.sql.snowflake.{SFQueryTest, SFTestSessionBase, SnowflakeSparkUtils}
+import org.apache.spark.sql.types.{LongType, StructType}
+
+class SnowflakeSparkUtilsSuite extends SFQueryTest with SFTestSessionBase {
```
sfc-gh-mrui, line 8: Use package `org.apache.spark.sql` because `JDBCRelation` is package private.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-mrui, line 113: If the pushdown is enabled, the spark plan statistic is added to `SnowflakeTelemetry.logs`, but it will be sent if any other kinds of messages is sent to snowflake lately.
In worst case, the pushdown is enabled, but the user doesn't not read/write from/to snowflake, all the spark plan statistic may be cached. To avoid OOM, SC cache at most 1000 messages.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -110,6 +120,7 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
         .format(SNOWFLAKE_SOURCE_NAME)
         .options(connectorOptionsNoTable)
         .option("dbtable", test_table)
+        .mode(SaveMode.Overwrite)
```
sfc-gh-ema, line 123: Impact of this change?

sfc-gh-mrui: It may be affected by the new test or something else.
I met test case failure because the save mode is `ErrorIfExists`. So, I update to use OverWrite mode.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -162,12 +173,197 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
...
+
+      val planStatisticMessages = messageBuffer
+        .filter(_.get("type").asText().equals("spark_plan_statistic"))
+        .filter { x =>
+          val planStatistics = x.get("data").get(TelemetryFieldNames.STATISTIC_INFO)
+          assert(planStatistics.isArray && planStatistics.size() > 0)
+          nodeContains(planStatistics, "LogicalRelation:SnowflakeRelation") &&
+            nodeContains(planStatistics, "SaveIntoDataSourceCommand:DefaultSource")
+        }
+      assert(planStatisticMessages.length == 1)
```
sfc-gh-ema, line 233: It seems there is a lot of duplicated code for every test case. Can we move the common logic into a method?

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-ema: If the cache is full, can we flush it? Otherwise it seems we'll miss the statistic. I wonder if just returning that will cause us to miss a lot of stats.

sfc-gh-mrui: SC can't flush it, because addSparkPlanStatistic() is called in pushdown analysis, there is no connection or sfOptions yet. It's possible to miss some plans.
In the case spark connector is used, it may not be possible because pushdown analysis is before read/write.

sfc-gh-ema: Can we try to flush it (with try/catch)? We should do our best to not miss messages. It seems like 1000 may not be a lot, since there can be many plans processed, and each plan can have a lot of nodes, and each node can have many expressions.

Another idea is to keep some counter of how many messages were missed/skipped and push that whenever we send the telemetry messages. That way we can at least have some idea if we are missing some messages and we can react in future releases.

sfc-gh-mrui: We have closed the connection once SC finished the query. I don't think we need to cache a connection for this purpose.
IMO, 1000 is a pretty big number because we only process Plan with `ReturnAnswer` on the top. No matter how complex a DataFrame is, only one message is generated. One `ReturnAnswer` is for one action.
I also think it is not necessary to track the missed_count, because it is low possibility that SC is used  after 1000+ non-spark-connector related actions.
The real world should be:
1. They don't use SC at all.
2. They use the SC a lot

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
sfc-gh-mrui: it can be extracted to a separate function. But I prefer not to do that. This is a test case, I think the easy-test-case-failure analysis is more important than code duplication.

src/it/scala/net/snowflake/spark/snowflake/TruncateTableSuite.scala
```diff
@@ -682,20 +682,30 @@ class TruncateTableSuite extends IntegrationSuiteBase {
...
-    // Check Utils.getLastSelectQueryId
-    assert(Utils.getLastSelect.contains(readQuery))
-    if (params.useCopyUnload) {
-      assert(getQueryTextFromHistory(Utils.getLastSelectQueryId).replaceAll("\\s+", "")
-        .contains(Utils.getLastSelect.replaceAll("\\s+", "")))
-    } else {
-      assert(getQueryTextFromHistory(Utils.getLastSelectQueryId).replaceAll("\\s+", "")
-        == Utils.getLastSelect.replaceAll("\\s+", ""))
-    }
+    try {
```
sfc-gh-mrui, line 685: Fix one flaky test issue.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-ema: Ok, I see now that we add this only for `ReturnAnswer`. That should be ok.

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -105,6 +105,28 @@ object SnowflakeTelemetry {
...
+          return
+        }
+      }
+
+      val metric: ObjectNode = mapper.createObjectNode()
+      val arrayNode = metric.putArray(TelemetryFieldNames.STATISTIC_INFO)
```
sfc-gh-bli, line 119: why array?

sfc-gh-bli: could you write a comment to show an example of message?

sfc-gh-mrui: There is a sample message in PR description.

sfc-gh-bli (approved):
LGTM

PR #451, opened 2022-08-11
SNOW-643250 Add spark dependencies to SPARK_CLIENT_INFO

sfc-gh-mrui:
SNOW-643250 Add spark dependencies to SPARK_CLIENT_INFO
Refer to chapter 6 for design discussion.
https://docs.google.com/document/d/1QzYH_1qCFOzu2a_ZoyRj27RzMDhVxU5mZMFNEX7BHeA/edit#heading=h.beye7n5os456

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
@@ -240,16 +240,18 @@ trait IntegrationEnv
     jdbcUpdate("alter session set timezone='UTC'")
 
     // Use fewer partitions to make tests faster
-    sparkSession = SparkSession.builder
-      .master("local")
-      .appName("SnowflakeSourceSuite")
-      .config("spark.sql.shuffle.partitions", "6")
-      // "spark.sql.legacy.timeParserPolicy = LEGACY" is added to allow
-      // spark 3.0 to support legacy conversion for unix_timestamp().
-      // It may not be necessary for spark 2.X.
-      .config("spark.sql.legacy.timeParserPolicy", "LEGACY")
-      .getOrCreate()
+    sparkSession = createDefaultSparkSession
   }
+
+  protected def createDefaultSparkSession = SparkSession.builder
+    .master("local")
+    .appName("SnowflakeSourceSuite")
+    .config("spark.sql.shuffle.partitions", "6")
+    // "spark.sql.legacy.timeParserPolicy = LEGACY" is added to allow
+    // spark 3.0 to support legacy conversion for unix_timestamp().
+    // It may not be necessary for spark 2.X.
+    .config("spark.sql.legacy.timeParserPolicy", "LEGACY")
+    .getOrCreate()
 
   override def afterAll(): Unit = {
     try {
--- a/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
@@ -10,7 +10,7 @@ import org.apache.spark.SparkEnv
 import org.apache.spark.ml.linalg.{Vector, Vectors}
 import org.apache.spark.ml.stat.ChiSquareTest
 import org.apache.spark.sql.functions._
-import org.apache.spark.sql.SaveMode
+import org.apache.spark.sql.{SaveMode, SparkSession}
 
 import scala.collection.mutable
 import scala.reflect.io.Directory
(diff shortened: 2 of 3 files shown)


src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -67,13 +67,74 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
...
+      sparkSession = SparkSession.builder
+        .master("local")
+        .appName("SnowflakeSourceSuite")
+        .config("spark.sql.shuffle.partitions", "6")
+        // "spark.sql.legacy.timeParserPolicy = LEGACY" is added to allow
+        // spark 3.0 to support legacy conversion for unix_timestamp().
+        // It may not be necessary for spark 2.X.
+        .config("spark.sql.legacy.timeParserPolicy", "LEGACY")
+        .config("spark.files", pythonTestFile)
+        .config("spark.archives", dummyArchiveFile)
```
sfc-gh-mrui, line 100: RE-create a spark session with file and archive.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeTelemetrySuite.scala
```diff
@@ -67,13 +67,74 @@ class SnowflakeTelemetrySuite extends IntegrationSuiteBase {
...
+      df1.collect()
+
+      val clientInfoMessages = messageBuffer
+        .filter(_.get("type").asText().equals("spark_client_info"))
+      assert(clientInfoMessages.nonEmpty)
+      clientInfoMessages.foreach { x =>
+        val sparkDependencies = x.get("data").get(TelemetryFieldNames.DEPENDENCIES)
+        assert(sparkDependencies.isArray && sparkDependencies.size() == 2)
+        assert(nodeContains(sparkDependencies, pythonTestFile))
+        assert(nodeContains(sparkDependencies, dummyArchiveFile))
```
sfc-gh-mrui, line 127: The file and archive can be found in the telemetry message

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -93,6 +94,22 @@ object SnowflakeTelemetry {
     }
   }
 
+  private[snowflake] def getSparkDependencies: Seq[String] = {
+    try {
+      val activeSparkSession = SparkSession.getActiveSession
+      if (activeSparkSession.nonEmpty) {
+        val sparkContext = activeSparkSession.get.sparkContext
+        (sparkContext.files ++ sparkContext.archives).distinct
```
sfc-gh-ema, line 102: Is `sparkContext.archives` only going to pick up Python libs? And can you write a comment here about this?

sfc-gh-mrui: According to the document. It may be used for other purpose. https://spark.apache.org/docs/latest/configuration.html
``` 
Comma-separated list of archives to be extracted into the working directory of each executor. .jar, 
.tar.gz, .tgz and .zip are supported. You can specify the directory name to unpack via adding # after
 the file name to unpack, for example, file.zip#directory. This configuration is experimental.
```

sfc-gh-ema: So we may have some duplication in `dependencies` with `libraries`, right?

sfc-gh-mrui: logically, it is possible. But I think archive/file are mainly for non-jar files.
We can also retrieve jar files by `SparkContext.jars`, this may be duplicated to `libraries`.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -435,7 +454,12 @@ object SnowflakeTelemetry {
     "spark.sql.ansi.enabled",
     "spark.pyspark.driver.python",
     "spark.pyspark.python",
-    "spark.sql.session.timeZone"
+    "spark.sql.session.timeZone",
+    // If the users need to use Pandas DataFrame with Spark DataFrame,
+    // spark.sql.execution.arrow.pyspark.enabled needs to be set to "true"
```
sfc-gh-mrui, line 459: Add these 3 options to white list.

If users needs to use Pandas DataFrame, for example, convert a Spark DataFrame to a Pandas DataFrame,
They need to enable "spark.sql.execution.arrow.pyspark.enabled" . 
So I think we can add them.

src/it/scala/net/snowflake/spark/snowflake/IntegrationEnv.scala
```diff
@@ -240,17 +240,19 @@ trait IntegrationEnv
...
-      .config("spark.sql.shuffle.partitions", "6")
-      // "spark.sql.legacy.timeParserPolicy = LEGACY" is added to allow
-      // spark 3.0 to support legacy conversion for unix_timestamp().
-      // It may not be necessary for spark 2.X.
-      .config("spark.sql.legacy.timeParserPolicy", "LEGACY")
-      .getOrCreate()
+    sparkSession = createDefaultSparkSession
   }
 
+  protected def createDefaultSparkSession = SparkSession.builder
```
sfc-gh-bli, line 246: why move it out?
if necessary can we use `lazy protected val ...` to replace `protected var sparkSession: SparkSession = _`?

sfc-gh-mrui: In the unit test, I need to create a SparkSession with extra Spark options.
To create a SparkSession, I need to stop the current SparkSession first.
Extract this as a separate function so that I can create the default SparkSession after the test case done.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
sfc-gh-bli:
```scala
SparkSession.getActiveSession
  .map(_.sparkContext)
  .map(context => (context.files ++ context.archives).distinct)
  .getOrElse(Seq.empty)
```

sfc-gh-bli (approved):

sfc-gh-ema (approved):

PR #456, opened 2022-08-18
Test JDBC built from master branch

sfc-gh-mrui:


sfc-gh-mrui:
This is test only PR.

PR #467, opened 2022-10-17
SNOW-608021 Support S3 VPCE deployment

sfc-gh-mrui:
SNOW-608021 Support S3 VPCE deployment

1. Add a new parameter `S3_STAGE_VPCE_DNS_NAME`. If it is configured, SC will set it for the session.
2. For writing to snowflake, SC needs below changes:
    a. GS will send the stage endpoint for `put` command to JDBC. SC driver gets it from JDBC.
    b. SC driver passes it to SC executors
    c. When SC executor uploads data to the stage, it uses the stage endpoint for uploading.
3. For reading from snowflake, SC dons't need any change because JDBC is used to download the result chunks.

NOTE:
I can only test VPCE support manually because the test with it need VPN to `US West`.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -208,6 +208,16 @@ object Parameters {
     "internal_use_aws_region_url"
   )
 
+  // Session level parameter S3_STAGE_VPCE_DNS_NAME
+  val PARAM_S3_STAGE_VPCE_DNS_NAME: String = knownParam("s3_stage_vpce_dns_name")
+
+  // Internal option to support AWS stage endPoint. This may be used for AWS S3 FIPS and VPCE.
+  // by default, it is 'true'. This option may be removed without any notice in any time.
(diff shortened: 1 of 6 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -208,6 +208,16 @@ object Parameters {
     "internal_use_aws_region_url"
   )
 
+  // Session level parameter S3_STAGE_VPCE_DNS_NAME
+  val PARAM_S3_STAGE_VPCE_DNS_NAME: String = knownParam("s3_stage_vpce_dns_name")
```
sfc-gh-bli, line 212: why do we need prefix `PARAM_` here?

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: This is the name pattern we followed.

sfc-gh-ema (approved):

PR #470, opened 2022-11-10
SNOW-674428 Fix a NPE when sending telemetry messages.

sfc-gh-mrui:
SNOW-674428 Fix a NPE when sending telemetry messages.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
@@ -245,13 +245,15 @@ object SnowflakeTelemetry {
       logger.warn(s"Fail to send OOB Telemetry message: ${th.getMessage}")
   }
 
-  def send(telemetry: Telemetry): Unit = {
+  def send(telemetry: Telemetry): Unit = if (telemetry != null) {
     var curLogs: List[(ObjectNode, Long)] = Nil
     this.synchronized {
       curLogs = logs
       logs = Nil
     }
     telemetryMessageSender.send(telemetry, curLogs)
+  } else {
+    logger.warn(s"Don't send Telemetry message because Telemetry is null")
   }
 
   /**
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
@@ -245,6 +245,11 @@ object SnowflakeTelemetry {
       logger.warn(s"Fail to send OOB Telemetry message: ${th.getMessage}")
   }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -245,13 +245,15 @@ object SnowflakeTelemetry {
...
-  def send(telemetry: Telemetry): Unit = {
+  def send(telemetry: Telemetry): Unit = if (telemetry != null) {
     var curLogs: List[(ObjectNode, Long)] = Nil
     this.synchronized {
       curLogs = logs
       logs = Nil
     }
     telemetryMessageSender.send(telemetry, curLogs)
+  } else {
+    logger.warn(s"Don't send Telemetry message because Telemetry is null")
```
sfc-gh-ema, line 256:
```suggestion
    logger.warn(s"Telemetry message not sent because Telemetry object is null")
```

sfc-gh-ema (approved):

src/main/scala/net/snowflake/spark/snowflake/SnowflakeTelemetry.scala
```diff
@@ -245,13 +245,15 @@ object SnowflakeTelemetry {
...
-  def send(telemetry: Telemetry): Unit = {
+  def send(telemetry: Telemetry): Unit = if (telemetry != null) {
     var curLogs: List[(ObjectNode, Long)] = Nil
     this.synchronized {
       curLogs = logs
       logs = Nil
     }
     telemetryMessageSender.send(telemetry, curLogs)
+  } else {
+    logger.warn("Telemetry message not sent because Telemetry client is null")
```
sfc-gh-bli, line 256: This warn will be prompted for every telemetry message. Should just check one time when initializing telemetry service?

sfc-gh-bli:
```
lazy val skipTelemetry = {
   val skip = telemetry != null
   if (skip) warn
   skip
}
if (! skipTelemetry) {
....
}
```

sfc-gh-mrui: Changed like below:
```
  private lazy val telemetryClientNullWarningSent = {
    logger.warn("Telemetry message not sent because Telemetry client is null")
    true
  }

  def send(telemetry: Telemetry): Unit = if (telemetry != null) {
    var curLogs: List[(ObjectNode, Long)] = Nil
    this.synchronized {
      curLogs = logs
      logs = Nil
    }
    telemetryMessageSender.send(telemetry, curLogs)
  } else {
    telemetryClientNullWarningSent
  }
```

PR #471, opened 2022-11-10
SNOW-664248 Support proxy protocol for S3

sfc-gh-mrui:
SNOW-664248 Support proxy protocol for S3

Add new option `proxy_protocol` so that customer can configure it. valid values are `http` and `https`.

NOTE: This option only support snowflake AWS account. It is not supported for Azure and GCP.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -131,6 +131,7 @@ object Parameters {
 
   // Proxy related info
   val PARAM_USE_PROXY: String = knownParam("use_proxy")
+  val PARAM_PROXY_PROTOCOL: String = knownParam("proxy_protocol")
   val PARAM_PROXY_HOST: String = knownParam("proxy_host")
   val PARAM_PROXY_PORT: String = knownParam("proxy_port")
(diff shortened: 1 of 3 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -131,6 +131,7 @@ object Parameters {
 
   // Proxy related info
   val PARAM_USE_PROXY: String = knownParam("use_proxy")
+  val PARAM_PROXY_PROTOCOL: String = knownParam("proxy_protocol")
```
sfc-gh-bli, line 134: if it only works with AWS, should we name it `aws_proxy_protocol`?
Or do we have a plan to support azure and gcp later?

sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: This name is from JDBC.
Currently, it only support AWS. It may support Azure in the future. So it's better to use the same name.

sfc-gh-ema (approved):

PR #472, opened 2022-11-11
SNOW-692659 Add an option to support returning Long for the Decimal(X, 0) column.

sfc-gh-mrui:
SNOW-692659 Add an option to support returning Long for the Decimal(X, 0) column.

Add a new option: `treat_decimal_as_long` to support returning Long for the Decimal(X, 0) column.
WARNING: if the Decimal value is greater than max value of Long, conversion will happen.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -128,6 +128,11 @@ object Parameters {
   val PARAM_USE_AWS_MULTIPLE_PARTS_UPLOAD: String = knownParam(
     "use_aws_multiple_parts_upload"
   )
+  // Treat Decimal(X, 0) as Long when reading from Snowflake
+  // WARNING: if the Decimal value is greater than max value of Long, conversion will happen.
+  val PARAM_TREAT_DECIMAL_X_0_AS_INT: String = knownParam(
+    "treat_decimal_x_0_as_int"
+  )
 
   // Proxy related info
   val PARAM_USE_PROXY: String = knownParam("use_proxy")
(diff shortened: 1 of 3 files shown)


sfc-gh-bli (approved):
LGTM

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -128,6 +128,11 @@ object Parameters {
   val PARAM_USE_AWS_MULTIPLE_PARTS_UPLOAD: String = knownParam(
     "use_aws_multiple_parts_upload"
   )
+  // Treat Decimal(X, 0) as Long when reading from Snowflake
+  // WARNING: if the Decimal value is greater than max value of Long, conversion will happen.
+  val PARAM_TREAT_DECIMAL_X_0_AS_LONG: String = knownParam(
```
sfc-gh-ema, line 133: Can we just call this `treat_decimal_as_long`? We don't need the X.

sfc-gh-mrui: We can. I thought it doesn't affect Decimal with scale != 0. It looks it is not necessary.
Will update it as `TREAT_DECIMAL_X_0_AS_LONG`.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -899,6 +904,9 @@ object Parameters {
     def useAwsMultiplePartsUpload: Boolean =
       isTrue(parameters(PARAM_USE_AWS_MULTIPLE_PARTS_UPLOAD))
 
+    def treadDecimalAsLong: Boolean =
```
sfc-gh-ema, line 907: typo: `treatDecimalAsLong`

sfc-gh-mrui: thanks!

sfc-gh-ema (approved):

PR #488, opened 2023-01-07
SNOW-717324 Support sharing JDBC connections

sfc-gh-mrui:
SNOW-717324 Support sharing JDBC connections

The design spec can be found at https://docs.google.com/document/d/150n9sG9yOCNjQzxcPItyRs7TVqnuT0DPUvTL-ivG8kg/edit#

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
@@ -224,6 +224,11 @@ object Parameters {
     "internal_support_aws_stage_end_point"
   )
 
+  // Support to share JDBC connection if possible
+  val PARAM_SUPPORT_SHARE_CONNECTION: String = knownParam(
+    "support_share_connection"
+  )
+
   val DEFAULT_S3_MAX_FILE_SIZE: String = (10 * 1000 * 1000).toString
   val MIN_S3_MAX_FILE_SIZE = 1000000
 
@@ -697,6 +702,9 @@ object Parameters {
     }
     def supportAWSStageEndPoint: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_SUPPORT_AWS_STAGE_END_POINT, "true"))
+    }
+    def supportShareConnection: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_SUPPORT_SHARE_CONNECTION, "true"))
     }
     def stagingTableNameRemoveQuotesOnly: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_STAGING_TABLE_NAME_REMOVE_QUOTES_ONLY, "false"))
--- a/src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
@@ -1,0 +1,290 @@
+package net.snowflake.spark.snowflake
+
+import net.snowflake.client.core.{SFBaseSession, SFSession, SFSessionProperty}
+import net.snowflake.client.jdbc.SnowflakeConnectionV1
+import net.snowflake.spark.snowflake.Parameters.MergedParameters
+import net.snowflake.spark.snowflake.Utils.JDBC_DRIVER
+import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry
+import org.slf4j.LoggerFactory
+
+import java.sql.{Connection, DriverManager, PreparedStatement, SQLException, Statement}
+import java.util.Properties
+import java.util.concurrent.atomic.AtomicLong
+import scala.collection.mutable
+
+/*
+ * This class is the key if a JDBC connection is cached for sharing.
+ */
+private[snowflake]
+sealed class ConnectionCacheKey(private val parameters: MergedParameters) {
+  ServerConnection.serverConnectionCount.incrementAndGet()
+  private val PARAMETERS_NOT_AFFECT_SHARE_CONNECTION = Set(
+    // If it is SAVE, SC added overwrite internally, refer to DefaultSource.createRelation()
+    "overwrite",
+    Parameters.PARAM_POSTACTIONS,
+    Parameters.PARAM_PREACTIONS,
+    Parameters.PARAM_SF_QUERY,
+    Parameters.PARAM_SF_DBTABLE
+  )
+  private val connectionSharableCheckOptions =
+    parameters.parameters -- PARAMETERS_NOT_AFFECT_SHARE_CONNECTION
+
+  override def toString: String = "Connection Cache Key"
+
+  override def equals(obj: Any): Boolean = obj match {
+    case other: ConnectionCacheKey =>
+      connectionSharableCheckOptions.equals(other.connectionSharableCheckOptions)
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    connectionSharableCheckOptions.toArray.sorted.mkString(",").hashCode
+
+  private[snowflake] def isQueryInWhiteList(query: String) = {
+    // CREATE [ OR REPLACE ]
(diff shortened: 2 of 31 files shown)


src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -224,6 +224,11 @@ object Parameters {
     "internal_support_aws_stage_end_point"
   )
 
+  // Support to share JDBC connection if possible
```
sfc-gh-mrui, line 227: We introduce 2 methods to disable shared connection:
1. Spark Connector option: `support_share_connection` ( default is true)
2. Static function: SparkConnectorContext.disableSharedConnection()

In most cases, customer may create a sfOptions map and use it in everywhere.
I suppose method 1 is used in most cases.

Method 2 can be used if user can't update sfOptions easily.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -0,0 +1,290 @@
...
+
+  override def toString: String = "Connection Cache Key"
+
+  override def equals(obj: Any): Boolean = obj match {
+    case other: ConnectionCacheKey =>
+      connectionSharableCheckOptions.equals(other.connectionSharableCheckOptions)
+    case _ => false
+  }
+
+  override def hashCode(): Int =
```
sfc-gh-mrui, line 40: We uses a hash map:  to cache the JDBC connection:
`cachedJdbcConnections = mutable.Map[ConnectionCacheKey, Connection]()`

So, override `equals()` and `hashCode()` accordingy.
For `hashCode()`, we order the entries in `connectionSharableCheckOptions`, and make a string, and use the result String's hash code.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -0,0 +1,290 @@
...
+  private[snowflake] def setSupportSharingJDBCConnection(enabled: Boolean): Unit =
+    supportSharingJDBCConnection = enabled
+
+  def apply(jdbcConnection: Connection, enableCache: Boolean): ServerConnection =
+    new ServerConnection(jdbcConnection, enableCache)
+
+  def apply(jdbcConnection: Connection): ServerConnection =
+    new ServerConnection(jdbcConnection, true)
+
+  private val cachedJdbcConnections = mutable.Map[ConnectionCacheKey, Connection]()
```
sfc-gh-mrui, line 96: This is the JDBC connection cache Map.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -0,0 +1,290 @@
...
+
+  def closeAllCachedConnections: Unit = synchronized {
```
sfc-gh-mrui, line 98: This is called by the Application End listener.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -0,0 +1,290 @@
...
+    if (newConnection) {
+      val extraValues = Map(TelemetryClientInfoFields.SFURL -> parameters.sfURL,
+        TelemetryClientInfoFields.SHARED -> enableCache.toString)
+      SnowflakeTelemetry.sendClientInfoTelemetry(extraValues, serverConnection)
+    }
+
+    serverConnection
+  }
+
+  private def createJDBCConnection(params: MergedParameters): Connection = {
```
sfc-gh-mrui, line 133: This function is moved from `DefaultJDBCWrapper.getConnector()`. The only extra change is to move below to `getServerConnection()`
```
    // Send client info telemetry message.
    if (newConnection) {
      val extraValues = Map(TelemetryClientInfoFields.SFURL -> parameters.sfURL,
        TelemetryClientInfoFields.SHARED -> enableCache.toString)
      SnowflakeTelemetry.sendClientInfoTelemetry(extraValues, serverConnection)
    }
```

src/main/scala/net/snowflake/spark/snowflake/SnowflakeConnectorUtils.scala
```diff
@@ -113,6 +113,22 @@ object SnowflakeConnectorUtils {
       throw ex
     }
   }
+
```
sfc-gh-mrui, line 116: Added 2 Utility methods to disable/enable shared connection.
They are added to `SnowflakeConnectorUtils` because enablePushdown() is in it.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -112,141 +107,8 @@ private[snowflake] class JDBCWrapper {
...
-    }
-
-    // Send client info telemetry message.
-    val extraValues = Map(TelemetryClientInfoFields.SFURL -> sfURL)
-    SnowflakeTelemetry.sendClientInfoTelemetry(extraValues, conn)
-
-    conn
-  }
+  def getConnector(params: MergedParameters): ServerConnection =
+    ServerConnection.getServerConnection(params)
```
sfc-gh-mrui, line 111: The actual JDBC connection creation is moved to `ServerConnection.createJDBCConnection()`

src/main/scala/net/snowflake/spark/snowflake/SnowflakeJDBCWrapper.scala
```diff
@@ -331,7 +193,7 @@ private[snowflake] class JDBCWrapper {
     })
   }
 
-  def executePreparedInterruptibly(conn: Connection, sql: String): Boolean = {
+  def executePreparedInterruptibly(conn: ServerConnection, sql: String): Boolean = {
```
sfc-gh-mrui, line 196: In production code, use `ServerConnection` instead of `Connection` and `SnowflakeConnectionV1`.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -698,6 +703,9 @@ object Parameters {
     def supportAWSStageEndPoint: Boolean = {
       isTrue(parameters.getOrElse(PARAM_INTERNAL_SUPPORT_AWS_STAGE_END_POINT, "true"))
     }
+    def supportShareConnection: Boolean = {
+      isTrue(parameters.getOrElse(PARAM_SUPPORT_SHARE_CONNECTION, "true"))
```
sfc-gh-bli, line 707: Enable it by default will be a behavior change. I think we should disable it by default.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
sfc-gh-bli: Why not session ID?

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
sfc-gh-mrui: I prefer to regard this is a performance optimization instead of a behaviour change.
With this change, we expect the customer didn't see any change except it is faster.

It's like `ENABLE_FIX_<JIRA>` in snowflake. It is introduced just in order to give the user a way to disable in case of regression.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
sfc-gh-mrui: `ConnectionCacheKey` is created from `MergedParameters`. Two JDBC connections created for the same `MergedParameters` must has different session id.

sfc-gh-bli: No matter what parameters are, two JDBC connections have different session ID, even they have same connection parameters.
My idea is, we can reuse Snowflake's connection ID, instead of generating a new one.

sfc-gh-bli (approved):
LGTM

sfc-gh-ema (approved):

PR #492, opened 2023-03-08
Support micro seconds precisison during copy unload

arthurli1126:
Snowflake connector uses [SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) (legacy) for data time parsing, internally, it only supports milliseconds. For instance, for string `2023-03-01 07:54:56.191173` it would consider it carries 191173 milliseconds so it will add 
191000 / 1000 / 60 = 3 mins 11s and put 173 microseconds to milliseconds filed: 
2023-03-01 07:58:07.173000.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/Conversions.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/Conversions.scala
@@ -41,15 +41,18 @@ private[snowflake] object Conversions {
   // Note - we use a pattern with timezone in the beginning, to make sure
   // parsing with PATTERN_NTZ fails for PATTERN_TZLTZ strings.
   // Note - for JDK 1.6, we use Z ipo XX for SimpleDateFormat
+  // Because simpleDateFormat only support milliseconds,
+  // we need to refactor this and handle nano seconds field separately
   private val PATTERN_TZLTZ =
     if (System.getProperty("java.version").startsWith("1.6.")) {
-      "Z yyyy-MM-dd HH:mm:ss.SSS"
+      "Z yyyy-MM-dd HH:mm:ss."
     } else {
-      "XX yyyy-MM-dd HH:mm:ss.SSS"
+      "XX yyyy-MM-dd HH:mm:ss."
     }
 
   // For NTZ, Snowflake serializes w/o timezone
-  private val PATTERN_NTZ = "yyyy-MM-dd HH:mm:ss.SSS"
+  // and handle nano seconds field separately during parsing
+  private val PATTERN_NTZ = "yyyy-MM-dd HH:mm:ss."
 
   // For DATE, simple ISO format
   private val PATTERN_DATE = "yyyy-MM-dd"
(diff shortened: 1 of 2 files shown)


src/test/scala/net/snowflake/spark/snowflake/ConversionsSuite.scala
```diff
@@ -193,4 +193,88 @@ class ConversionsSuite extends FunSuite {
 
     assert(expect == result.toString())
   }
+
+  test("Data with micro-seconds and nano-seconds precision should be correctly converted"){
```
sfc-gh-mrui, line 197: This is a unit test. It proves that it parse timestamp with micro/nano-seconds.
Could you please add an integration test  (AKA end-to-end test)?

sfc-gh-mrui (changes requested):
https://github.com/snowflakedb/spark-snowflake/blob/master/src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala#L711 is the test case for new Arrow Format read (non-COPY-UNLOAD). 
It is disabled for COPY UNLOAD. It will be great if you can make it work.

src/it/scala/net/snowflake/spark/snowflake/SnowflakeResultSetRDDSuite.scala
```diff
@@ -710,111 +710,102 @@ class SnowflakeResultSetRDDSuite extends IntegrationSuiteBase {
 
   test("testTimestamp") {
     setupTimestampTable
-    // COPY UNLOAD can't be run because it only supports millisecond(0.001s).
-    if (!params.useCopyUnload) {
-      val result = sparkSession.sql("select * from test_table_timestamp")
+    val result = sparkSession.sql("select * from test_table_timestamp")
```
sfc-gh-mrui, line 713: This test case failed.

arthurli1126: @Mingli-Rui Yea sorry, haven't finished the change yet, should've mentioned this. Will push another commit this week.

arthurli1126:
@sfc-gh-mrui The test cases has been fixed please review. Thanks!

src/main/scala/net/snowflake/spark/snowflake/Conversions.scala
```diff
@@ -193,7 +196,23 @@ private[snowflake] object Conversions {
     * Parse a string exported from a Snowflake TIMESTAMP column
     */
   private def parseTimestamp(s: String, isInternalRow: Boolean): Any = {
+    // Need to handle the nano seconds filed separately
```
sfc-gh-mrui, line 199: Could you please add an internal parameter to enable/disable the change? It's enabled by default.
If some users is broken, they can disable the fix as a workaround.
For example, https://github.com/snowflakedb/spark-snowflake/blob/master/src/main/scala/net/snowflake/spark/snowflake/Parameters.scala#LL163C14-L163C14

arthurli1126: hmm.. do you mean to have a parameter such as `internal_support_micro_second_during_unload` ?  IMO, we should always try to support micro second level precision since by default direct JDBC supports this precision. It would make this part of the code confusing if we have two code paths and two set of timestamp patterns..

sfc-gh-mrui: I agree users need to use this fix. This is why we can set the internal parameter as true by default.
This is the internal policy to introduce a parameter to disable the fix if possible.
The internal parameter can be removed later. So the code will be clean up accordingly.

arthurli1126: Thanks for the suggestion. Will push a new commit to add this parameter.

sfc-gh-mrui:
@arthurli1126 Let's me know if you are ready for review. So far, below test case failed.
```
23/03/27 19:12:32 INFO DAGScheduler: Job 109 finished: collect at DataTypesIntegrationSuite.scala:244, took 0.006617 s
[info] - filter on timestamp column *** FAILED ***
```

arthurli1126:
@sfc-gh-mrui  Yea the tests pass in my local and with instacart's snowflake account(based on aws), I realized the test only failed with azure and gcp copy unload so wondering if there's anything snowflake does differently when unload to gcp/azure. However, I'm not able to verify it due to my company only has aws deployment. Is this something we can help me with or can you provide a testing temp GCP deployment account?

arthurli1126:
Looks I can sign up for a trial account with snowflake on GCP. Let me also try that.

arthurli1126:
Sorry about the long break finally got chance to look into this again. For workflows in different envs, does it need approval from your side to run? @sfc-gh-mrui

arthurli1126:
Hi @sfc-gh-mrui, I was wondering if there's a way to trigger the test again? Thanks 🙏

PR #502, opened 2023-03-28
SNOW-763124 Support uploading files with down-scoped token for Gcs

sfc-gh-mrui:
SNOW-763124 Support uploading files with down-scoped token for Snowflake Gcs accounts

The GCS down-scoped token support needs both JDBC and SC to change.
The JDBC change has been included in JDBC 3.13.30. https://github.com/snowflakedb/snowflake-jdbc/pull/1316

This PR also includes:
1. Upgrade the JDBC and SC version to 3.13.30 and 2.11.3.
2. Enhance SC to support down-scoped token for GCP.

The design is https://docs.google.com/document/d/1MlGUyoFOAYEJx6skw2KZfduKp1EPQIJ2SFdsm3emOJ4/edit#

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
@@ -738,6 +738,7 @@ sealed trait CloudStorage {
           .setSnowflakeFileTransferMetadata(fileTransferMetadata.get)
           .setUploadStream(inStream)
           .setRequireCompress(compress)
+          .setDestFileName(fileName)
           .setOcspMode(OCSPMode.FAIL_OPEN)
           .setProxyProperties(proxyProperties)
           .build())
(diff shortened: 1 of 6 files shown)


src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -738,6 +738,7 @@ sealed trait CloudStorage {
           .setSnowflakeFileTransferMetadata(fileTransferMetadata.get)
           .setUploadStream(inStream)
           .setRequireCompress(compress)
+          .setDestFileName(fileName)
```
sfc-gh-mrui, line 741: For uploading multiple partitions, we only use ` PUT` command to get uploading one metadata, So we need to set `destFileName` for each file uploading. 
PS: `SnowflakeFileTransferAgent.uploadWithoutConnection()` has been enhanced to use the DestFileName.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -1769,18 +1778,29 @@ case class InternalGcsStorage(param: MergedParameters,
         new SFStatement(connection.getSfSession)
       ).getFileTransferMetadatas
         .asScala
-        .map(oneMetadata => result += oneMetadata)
+        .foreach(oneMetadata => {
+          if (!oneMetadata.isForOneFile) {
```
sfc-gh-mrui, line 1782: `isForOneFile()` is an existing API. It has been enhanced to return true if `down-scoped token` is used.

src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
```diff
@@ -1839,14 +1859,15 @@ case class InternalGcsStorage(param: MergedParameters,
...
+
     val endTime = System.currentTimeMillis()
     CloudStorageOperations.log.info(
       s"""${SnowflakeResultSetRDD.MASTER_LOG_PREFIX}:
          | Finish uploading data for ${data.getNumPartitions} partitions in
          | ${Utils.getTimeString(endTime - startTime)}.
          |""".stripMargin.filter(_ >= ' '))
-
-    fileUploadResults.collect().toList
+    result
```
sfc-gh-mrui, line 1870: The change for line 1862 - 1870 is optional for GCP support, but it is necessary to make logging time to be accurate.

The spark executors are done after `fileUploadResults.collect()` is returned. So, I change it before logging.

sfc-gh-ema (approved):

PR #510, opened 2023-05-09
SNOW-824475 Support Spark 3.4

sfc-gh-mrui:
SNOW-824475 Support Spark 3.4

The main changes for Spark 3.4 affects Spark Connector are:
1. Change the last argument for `Cast`
    - Old type: `ansiEnabled: Boolean`
    - New Type: `evalMode: EvalMode.Value`
    - Currently, there are 3 modes: `LEGACY`, `ANSI`, `TRY`
    - Support to pushdown, if the mode is `LEGACY`. It will make SC to be in consistent behaviour.
2. Added a `JoinHint` argument for `ScalarSubquery`.
    - Spark Connector only pushdown if joinCont is not empty. So the new parameter can be ignored.
3. `PromotePrecision` is removed from Spark 3.4
    - So SC only needs to remove the process for this pan node.
4. Added a new argument `ansiEnabled` for `Round`.
    - Support to pushdown, if the ansiEnabled is false. It will make SC to be in consistent behaviour.

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
@@ -334,15 +334,22 @@ class SimpleNewPushdownIntegrationSuite extends IntegrationSuiteBase {
     var result =
       sparkSession.sql(s"select o $operator p from df2 where o IS NOT NULL")
 
-    testPushdown(
+    val expectedAdditionQueries = Seq(
       s"""SELECT ( CAST ( ( "SUBQUERY_1"."O" $operator "SUBQUERY_1"."P" )
-                    |AS DECIMAL(38, 0) ) ) AS "SUBQUERY_2_COL_0" FROM
-                    |( SELECT * FROM ( SELECT * FROM ( $test_table2 ) AS
-                    |"SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0" WHERE
-                    |( "SUBQUERY_0"."O" IS NOT NULL ) ) AS "SUBQUERY_1"
+         |AS DECIMAL(38, 0) ) ) AS "SUBQUERY_2_COL_0" FROM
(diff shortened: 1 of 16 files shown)


sfc-gh-bli (approved):
LGTM

src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
```diff
@@ -370,15 +385,22 @@ class SimpleNewPushdownIntegrationSuite extends IntegrationSuiteBase {
...
+         |( "SUBQUERY_0"."O" IS NOT NULL ) ) AS "SUBQUERY_1"
       """.stripMargin,
-      result,
-      // Data in df2 (o, p) values(null, 1), (2, 2), (3, 2), (4, 3)
+      // From spark 3.4, the CAST operation is not presented in the plan
+      s"""SELECT ( ( "SUBQUERY_1"."O" $operator "SUBQUERY_1"."P" ) ) AS "SUBQUERY_2_COL_0" FROM
+         |( SELECT * FROM ( SELECT * FROM ( $test_table2 ) AS
+         |"SF_CONNECTOR_QUERY_ALIAS" ) AS "SUBQUERY_0" WHERE
+         |( "SUBQUERY_0"."O" IS NOT NULL ) ) AS "SUBQUERY_1"
+      """.stripMargin
```
sfc-gh-zli, line 400: If I understand correctly, we issue two SQLs now instead of one for this push down? Is that a performance issue?

sfc-gh-zli (approved):

src/it/scala/net/snowflake/spark/snowflake/SimpleNewPushdownIntegrationSuite.scala
sfc-gh-mrui: This is test-only and it doesn't cause any perf problem for production.
1. The generated query for different spark may be different.
2. To make the test case to work for different Spark versions, the `expectedMultiplyQueries` includes the query for spark 3.4 and previous version. If the query matches either one, the test case will succeed.

PR #520, opened 2023-08-25
Support spark.sql.datetime.java8API.enabled

tom-s-powell:
Support Spark's `spark.sql.datetime.java8API.enabled` property when `true`. This expects dates and timestamp to be `java.time.LocalDate` and `java.time.Instant` respectively rather than the `java.sql` equivalents.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/CSVConverter.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/CSVConverter.scala
@@ -17,7 +17,6 @@
 
 package net.snowflake.spark.snowflake
 
-import org.apache.spark.SparkConf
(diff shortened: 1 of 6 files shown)


tom-s-powell:
Any update on when we may be able to get this in?

tom-s-powell:
I've simplified this using `SQLConf.get.datetimeJava8ApiEnabled` rather than passing around `isJava8Time` flag, if that's more preferable as it results in a smaller diff.

tom-s-powell:
Any update on when this change could be reviewed?

tom-s-powell:
Going to close this given https://issues.apache.org/jira/browse/SPARK-38437.

PR #557, opened 2024-05-13
SNOW-1325562 Support Spark 3.5

sfc-gh-bli:
- Support Spark 3.5
- Remove cluster test
- Remove more pushdown related tests


src/it/scala/org/apache/spark/sql/SFDataFrameWindowFramesSuite.scala
```diff
@@ -167,42 +167,6 @@ class SFDataFrameWindowFramesSuite
         min("value").over(
           window.rangeBetween(Window.unboundedPreceding, Window.unboundedFollowing))),
       Row("non_numeric", "non_numeric") :: Nil)
-
```
sfc-gh-bli, line 170: No need to verify Spark version anymore.

sfc-gh-yuwang (approved):

PR #558, opened 2024-05-16
Add ABORT_DETACHED_QUERY parameter in session prologue query

urosstan-db:
When customer use async mode for query execution (async mode on JDBC driver) and he has ABORT_DETACHED_QUERY parameters set to true, then long running queries (longer than 5 minutes) being cancelled.

For more details, please take a look on https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query.

The current fix is to set session parameter to override user/account level parameter setting, while we still want to allow customer to pass desired value to options of spark read.

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
@@ -193,7 +193,7 @@ private[snowflake] case class SnowflakeRelation(
   ): RDD[T] = {
     val conn = DefaultJDBCWrapper.getConnector(params)
     try {
-      Utils.genPrologueSql(params).foreach(x => x.execute(bindVariableEnabled = false)(conn))
+      Utils.genPrologueSql(params).execute(bindVariableEnabled = false)(conn)
       Utils.executePreActions(DefaultJDBCWrapper, conn, params, params.table)
       Utils.setLastSelect(statement.toString)
       log.info(s"Now executing below command to read from snowflake:\n${statement.toString}")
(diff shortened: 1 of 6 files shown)


src/main/scala/net/snowflake/spark/snowflake/Utils.scala
```diff
@@ -398,7 +422,8 @@ object Utils {
          |date_output_format,
          |timestamp_ntz_output_format,
          |timestamp_ltz_output_format,
-         |timestamp_tz_output_format;
+         |timestamp_tz_output_format,
+         |ABORT_DETACHED_QUERY;
```
urosstan-db, line 426: This method is not used, but I added param for consistency.

src/main/scala/net/snowflake/spark/snowflake/SnowflakeRelation.scala
```diff
@@ -193,7 +193,7 @@ private[snowflake] case class SnowflakeRelation(
   ): RDD[T] = {
     val conn = DefaultJDBCWrapper.getConnector(params)
     try {
-      Utils.genPrologueSql(params).foreach(x => x.execute(bindVariableEnabled = false)(conn))
+      Utils.genPrologueSql(params).execute(bindVariableEnabled = false)(conn)
```
urosstan-db, line 196: We can also avoid merging timezone setting query and session parameter setting query into one query. But it would cost us some performances, so I wanted to merge them into one query.

sfc-gh-bli:
@urosstan-db thank you for your contribution.
We modified this PR and filed a new PR. That one has been merged.

PR #576, opened 2024-08-07
SNOW-1572335 Create DataFrame from Provided JDBC Connection

sfc-gh-bli:
SNOW-1572335 Create DataFrame from Provided JDBC Connection

### DIFF
--- a/src/it/scala/net/snowflake/spark/snowflake/ShareConnectionSuite.scala
+++ b/src/it/scala/net/snowflake/spark/snowflake/ShareConnectionSuite.scala
@@ -16,6 +16,8 @@
 
 package net.snowflake.spark.snowflake
 
+import net.snowflake.client.jdbc.SnowflakeConnectionV1
+
 import java.util.TimeZone
 import net.snowflake.spark.snowflake.Utils.SNOWFLAKE_SOURCE_NAME
 import net.snowflake.spark.snowflake.test.TestHook
@@ -422,5 +424,26 @@ class ShareConnectionSuite extends IntegrationSuiteBase {
     assert(result2.next())
     assert(result2.getString(1).equals("ABORT_DETACHED_QUERY"))
     assert(result2.getBoolean(2))
+  }
+
+  test("use provided connections") {
+    val conn1 =
+      ServerConnection
+        .getServerConnection(Parameters.mergeParameters(connectorOptions)).jdbcConnection
+    val id1 = ServerConnection.providedConnections.register(conn1)
+    assert(ServerConnection.providedConnections.hasConnectionID(id1))
+    val conn2 = ServerConnection.providedConnections.getConnection(id1).get
+    assert(conn1 == conn2)
+    // should remove after get
+    assert(!ServerConnection.providedConnections.hasConnectionID(id1))
+    val id2 = ServerConnection.providedConnections.register(conn1)
+    // should be same session id
+    assert(id2 == id1)
+//    sparkSession
+//      .read
+//      .format(SNOWFLAKE_SOURCE_NAME)
+//      .option(Parameters.PARAM_CONNECTION_ID, id2)
+//      .option("query", "select 1")
+//      .load().show()
   }
 }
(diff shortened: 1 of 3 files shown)


src/it/scala/net/snowflake/spark/snowflake/ShareConnectionSuite.scala
```diff
@@ -423,4 +425,26 @@ class ShareConnectionSuite extends IntegrationSuiteBase {
     assert(result2.getString(1).equals("ABORT_DETACHED_QUERY"))
     assert(result2.getBoolean(2))
   }
+
+  test("use provided connections") {
+    val conn1 =
+      ServerConnection
+        .getServerConnection(Parameters.mergeParameters(connectorOptions)).jdbcConnection
+    val id1 = ServerConnection.providedConnections.register(conn1)
```
sfc-gh-bli, line 433: This is an internal API can be accessed from snowflake package only. We may open it to public in feature if necessary.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -305,81 +307,87 @@ object Parameters {
       case (key, value) => (key.toLowerCase, value)
     }
 
+    if (!userParameters.contains(PARAM_CONNECTION_ID)) {
```
sfc-gh-bli, line 310: If connection id is presented in the parameters, all login related parameters will be optional.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -403,7 +411,11 @@ object Parameters {
       }
     }
 
-    MergedParameters(DEFAULT_PARAMETERS ++ userParameters)
+    val parametersFromSnowpark = userParameters.get(PARAM_CONNECTION_ID) match {
+      case Some(id) => ServerConnection.providedConnections.getParameters(id)
+      case _ => Map.empty
+    }
```
sfc-gh-bli, line 417: It will read `sf_url` and `sf_user` parameters from the provided JDBC connection.
Spark connector uses those values in the telemetry and log. We add those values here to avoid large amount of code change.

src/main/scala/net/snowflake/spark/snowflake/Parameters.scala
```diff
@@ -975,6 +987,10 @@ object Parameters {
       parameters.get(PARAM_JDBC_QUERY_RESULT_FORMAT)
     }
 
+    private[snowflake] def getConnectionId: Option[String] = {
+      parameters.get(PARAM_CONNECTION_ID)
+    }
```
sfc-gh-bli, line 992: It is internal too.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -109,19 +109,35 @@ private[snowflake] object ServerConnection {
   def getServerConnection(parameters: MergedParameters, useCache: Boolean = true)
   : ServerConnection = synchronized {
     val connectionCacheKey = new ConnectionCacheKey(parameters)
-    val enableCache = useCache && connectionCacheKey.isConnectionCacheSupported
+    // when use connection from Snowpark, always reuse cached the connection
+    // since the Snowpark connection in the staging area will be purged after get.
```
sfc-gh-bli, line 113: when connection id provided, Spark connector will enforce to reuse connections, since it can't create new connection with the given configurations.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -274,6 +290,36 @@ private[snowflake] object ServerConnection {
     jdbcConnectionCount.incrementAndGet()
     conn
   }
+
+  object providedConnections {
+    // use provided JDBC connection
+    private val providedConn = mutable.Map[String, Connection]()
```
sfc-gh-bli, line 296: This is the staging area to cache Snowflake connection. Since Spark API only accepts String values, we can't directly pass JDBC connection to Spark connector.
Instead of that, we cache the connection in the staging area and add the connection id into the parameter map. Therefore, connector can load the connection from the staging area.

src/main/scala/net/snowflake/spark/snowflake/ServerConnection.scala
```diff
@@ -274,6 +290,36 @@ private[snowflake] object ServerConnection {
...
+      providedConn.put(sessionId, conn)
+      sessionId
+    }
+    def hasConnectionID(id: String): Boolean =
+      providedConn.contains(id)
+
+    def getConnection(id: String): Option[Connection] = this.synchronized {
+      val result = providedConn.get(id)
+      if (result.isDefined) {
+        providedConn.remove(id)
```
sfc-gh-bli, line 308: Remove the cached the connection from staging area after get. Connector will use the regular shared connection feature to cache the provided connection.

sfc-gh-yuwang (approved):

PR #608, opened 2025-06-02
SNOW-2129112: Update jdbc version to 3.24.2

sfc-gh-wshangguan:
Update JDBC version to 3.24.2 to include a JDBC TrustManager fix.


sfc-gh-igarish (approved):
LGTM. But customer is using 2.12. Please double check.

PR #613, opened 2025-08-05
SNOW-2249717: Force Proxy Basic Auth in S3 Client

sfc-gh-wfateem:
We explicitly set BASIC as the authentication method for Proxy connections made by the S3 client when a proxy user and password are provided. Previously, the AWS S3 client could potentially attempt to connect using other methods if those were provided by the Proxy server in the Proxy-Authenticate headers


sfc-gh-wshangguan (approved):
LGTM.

PR #619, opened 2025-08-20
SNOW-2273060: Change to artifacts autorelease

sfc-gh-wshangguan:


deploy.sh
```diff
@@ -82,8 +82,7 @@ if [ "$PUBLISH" = true ]; then
   sbt +publishSigned
   # Upload to Central Portal and auto-release
   sbt sonaUpload
-  # TODO: should be do auto-release?
-  # sbt sonaRelease
+  sbt sonaRelease
```
sfc-gh-wshangguan, line 85: Currently we need release team to manually publish artifacts from the maven. With the change, the publish step becomes automatic. No need for human to do it manually, which aligns with many other Snowflake client releases

sfc-gh-saroskar (approved):

PR #645, opened 2026-04-09
Make S3A credentials test tolerate more error message variants

sfc-gh-helmeleegy:
Make S3A credentials test tolerate more error message variants


sfc-gh-bli (approved):
LGTM

PR #652, opened 2026-05-07
NOSNOW: Revert Spark4 work for the upcoming 3.1.9 release

sfc-gh-helmeleegy:
Spark 4 work is planned to be restored for a followup (BCR) release (after 3.1.9).


sfc-gh-bli (approved):
LGTM

codealchemy:
Process question from an observer here: could the relevant commits be cherry-picked to a (stable) version branch instead of reverts (and reverts of reverts) on `master? 

(perhaps a documented release process would be helpful, otherwise I just noticed the `spark_<version>` branches here and though it might be helpful for myself/others to ask).

sfc-gh-helmeleegy:
> Process question from an observer here: could the relevant commits be cherry-picked to a (stable) version branch instead of reverts (and reverts of reverts) on `master?
> 
> (perhaps a documented release process would be helpful, otherwise I just noticed the `spark_<version>` branches here and though it might be helpful for myself/others to ask).

Thanks for your question, and for the documentation note! We're adopting a cross-build model, but before our next major release, we wanted to have one more 3.1.x patch. Then we plan to ship cross-build + Spark 4 + JDBC 4 together. The revert/restore operations are meant to keep master as the source of truth.

PR #653, opened 2026-05-08
NOSNOW: Prevent silent data loss when speculative tasks are killed during stage upload

sfc-gh-helmeleegy:
## Summary
Under `spark.speculation=true`, a killed speculative task could silently
overwrite a successful sibling task's staged file. The truncated file is
still a valid GZIP/CSV, so `COPY INTO` completes without error and the
table ends up with fewer rows than the source — no exception, no log
warning visible on the driver. A user reported 100K+ rows lost in a
single job.

Root cause is in `CloudStorageOperations.doUploadPartition`: when Spark
interrupts a task and the `Thread.interrupted` flag has been cleared by
an intervening I/O handler, the unconditional `uploadStream.close()` in
the `finally` block still runs and commits the partial data via
`putObject` (single-shot path) or `completeMultipartUpload` (multipart
path), overwriting the good object that another attempt of the same
partition already wrote.

## Changes
1. **Skip `close()` on the failure path.** A `success` flag plus
   `TaskContext.isInterrupted()` guard the commit. If either signals
   trouble, the partial upload is abandoned.
2. **Abort multipart uploads explicitly.** For partitions large enough
   to trigger `S3UploadOutputStream`'s multipart path, parts have already
   been uploaded by the time the kill is observed. Skipping `close()`
   alone would leak the in-progress multipart upload (cost until the
   bucket's `AbortIncompleteMultipartUpload` lifecycle rule, if any,
   cleans them up). New `S3UploadOutputStream.abort()` issues
   `AbortMultipartUploadRequest` explicitly and is wired into the
   discard branch.
3. **Unit tests.** `S3UploadOutputStreamAbortSuite` exercises four
   `abort()` invariants directly against `AbstractAmazonS3`: no commit
   pre-write, no commit in the single-shot path, explicit
   `AbortMultipartUpload` (and no commit) when parts were already
   uploaded, and cleanup-failure swallowing.

## Credit
Root-cause analysis and the initial `success` flag + `TaskContext.isInterrupted()`
fix are by @BOOTMGR in #648. We're landing the work here so we can also
add the multipart-abort path and unit-test coverage. Closes #648.

Co-Authored-By: Harsh Panchal <panchal.harsh18@gmail.com>

### DIFF
--- a/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
+++ b/src/main/scala/net/snowflake/spark/snowflake/io/CloudStorageOperations.scala
@@ -739,7 +739,8 @@ sealed trait CloudStorage {
         }
         success = true
       } finally {
-        if (success) {
+        val isInterrupted = TaskContext.get() != null && TaskContext.get().isInterrupted()
+        if (success && !isInterrupted) {
           if (storageInfo.isDefined) {
             uploadStream.close()
           } else {
@@ -770,6 +771,10 @@ sealed trait CloudStorage {
                  | ${Utils.getTimeString(endTime - startTime)}
                  |""".stripMargin.filter(_ >= ' ')
           }
+        } else {
+          CloudStorageOperations.log.warn(
+            s"Discarding partial file $fileName. success=$success, isInterrupted=$isInterrupted"
+          )
         }
       }
     } else {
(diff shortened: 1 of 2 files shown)


sfc-gh-bli (approved):
LGTM

PR #656, opened 2026-05-28
NOSNOW: Upgrade Snowflake JDBC driver to 4.x

sfc-gh-helmeleegy:
Upgrade `snowflake-jdbc` from 3.28.0 to 4.0.2. JDBC 4.x no longer ships shaded AWS / Azure SDKs, so the connector now depends on them directly and relocates them into the released jar via `sbt-assembly`. The user-visible classpath is unchanged — only the JDBC driver upgrade is observable to customers.
This is a re-implementation of #647 on top of current `master` (which has since absorbed cross-build, Spark 4, and other changes).
**This is a BCR**: drops runtime support for JDBC 3.x. 

Co-authored-by: Bing Li <bing.li@snowflake.com>

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,6 +17,8 @@
 import java.io.File
 
 import scala.util.Properties
+import scala.xml.{Elem, Node => XmlNode, NodeSeq => XmlNodeSeq}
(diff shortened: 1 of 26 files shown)


sfc-gh-bli (approved):
LGTM

PR #658, opened 2026-06-18
NOSNOW: clean publish staging between Spark version iterations in deploy.sh

sfc-gh-helmeleegy:
Fixes the Central Portal duplicate-component validation failure that
blocked the spark_4.0 cell of the v3.2.0 release. Without per-iteration
cleanup, sonaUpload bundles previously-released spark_3.5 artifacts
together with the current iteration's spark_4.0 / spark_4.1 artifacts,
which Central Portal rejects as already existing on Maven Central.

Mirrors the existing rm -rf cleanup in the non-publish branch of the
script.


sfc-gh-bli (approved):
LGTM

PR #659, opened 2026-06-22
NOSNOW: clean the correct sona staging dir between Spark iterations

sfc-gh-helmeleegy:
The previous [PR](https://github.com/snowflakedb/spark-snowflake/pull/658) didn't fully fix the staging cleanup issue. I reproduced the issue locally (a mixed 3.5+4.0 bundle was incorrectly left in staging), and then verified that with the current fix, cleanup is done correctly (only 4.0 is left), which should unblock the jenkins publishing job.


sfc-gh-bli (approved):
LGTM

PR #660, opened 2026-06-22
NOSNOW: publish Spark 4 without the + cross-build prefix

sfc-gh-helmeleegy:
Spark 4.x sets crossScalaVersions to empty (Scala 2.13 only), so `+publishSigned` iterates over nothing and stages no artifacts. With the staging dir now correctly cleaned between iterations (#659), the empty publish surfaced as empty bundles for spark_4.0 / spark_4.1, which Maven Central Portal rejected (MavenPushSparkConnector). Previously the empty publish was masked by leftover spark_3.5 artifacts in staging.

Use +publishSigned only for Spark 3.x (cross-builds Scala 2.12 + 2.13); publish the single default Scala 2.13 for Spark 4.x without +.

Extend DeployScriptSuite to assert Spark 4 uses a non-+ publishSigned.


sfc-gh-bli (approved):
LGTM

PR #661, opened 2026-07-07
Restrict FallbackCatalog option merge to immutable Spark configuration

sfc-gh-helmeleegy:
Only immutable, connector-owned Spark configuration is merged into FallbackCatalog options. Runtime/session SQLConf entries are no longer folded into the catalog option map, so per-session values cannot override trusted catalog settings. Adds SnowflakeFallbackCatalogTest coverage.


sfc-gh-bli (approved):
LGTM

PR #662, opened 2026-07-07
Register connector secret options for Spark log redaction

sfc-gh-helmeleegy:
Registers the connector's sensitive DataSource options with Spark's redaction machinery so they are masked in logs, plan output, and the Spark UI. Applied to both the Spark 3.5 and 4.x source trees. Adds RedactionRegistrationSuite coverage.


sfc-gh-bli (approved):
LGTM

PR #663, opened 2026-07-07
Escape embedded quotes in identifier quoting helpers

sfc-gh-helmeleegy:
The identifier quoting helpers in Utils now escape embedded double quotes when wrapping identifiers, so values containing quote characters cannot break out of the quoted identifier context. Adds UtilsQuotingSecuritySuite coverage.


sfc-gh-bli (approved):
LGTM

PR #664, opened 2026-07-07
Escape SQL metacharacters when building CREATE STAGE

sfc-gh-helmeleegy:
CREATE STAGE construction now escapes SQL metacharacters in interpolated values so option/path inputs cannot alter the generated statement. Adds CreateStageEscapingSecuritySuite coverage.


sfc-gh-bli (approved):
LGTM

PR #665, opened 2026-07-07
Validate OAuth token request URL scheme and host

sfc-gh-helmeleegy:
The OAuth token request URL is validated for scheme and host before use, so a crafted token-request endpoint cannot redirect token requests to an attacker-controlled destination. Adds OAuthUrlValidationSecuritySuite coverage.


sfc-gh-bli (approved):
LGTM

